PR #62, opened 2018-01-08
HRRR grid

scotthavens:
Reading in HRRR model results into SMRF.


micah-prime:
This is now merged into develop. It should be merged into master with the upcoming 0.5.0 bump.

micahjohnson150:
The PR request was meant for develop. It was merged into develop prior to editing the base branch, now that it is in we cannot change the base branch since there is no new changes.

TLDR this was completely merged into develop where we wanted it.

Issue #87, opened 2019-03-27
Cloud factor at night


scotthavens:
The cloud factor at night was not being calculated properly with HRRR input data.The dataframe is being set to 0 initially and this isn't changed, so it's fully cloud coverage every night.

Hedrick-ARS:
REALLY!!! SHIT!

Could this be why we are coming up short for our storages?

scotthavens:
That is my thinking at the moment. Going to test today.

PR #88, opened 2019-03-27
Revamp the HRRR cloud factor calculation. Two major changes, first tw…

scotthavens:
…o stream is ran for each grid cell if the solar is greater than zero. This has the potential to make a little slower data load. Second, the HRRR solar dataframe and the two stream data frame are used to calculate the cloud factor. The NaN values at night are interpolated between the days. Fixes issue #87

### DIFF
--- a/smrf/envphys/radiation.py
+++ b/smrf/envphys/radiation.py
@@ -994,24 +994,44 @@ def get_hrrr_cloud(df_solar, df_meta, logger, lat, lon):
     clms = df_solar.columns
     dates = df_solar.index.values[:]
 
-    # find basin solar at top of atmosphere for each data
+    # find each cell solar at top of atmosphere for each date
     basin_sol = np.zeros(len(dates))
     basin_sol[:] = np.nan
-    for idt, dt in enumerate(dates):
-        # get solar using twostream
+
+
+    cs_solar = df_solar.copy()
+    for dt, row in cs_solar.iterrows():
         dtt = pd.to_datetime(dt)
-        basin_sol[idt] =  model_solar(dtt, lat, lon)
+        for ix,value in row.iteritems():
+            # get solar using twostream only if the sun is up
+            if value > 0:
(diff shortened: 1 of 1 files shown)


micahjohnson150:
Passed my python 3.6 tests

micah-prime:
Here is a quick summary of this change with a little documentation

## Description

There was an issue with the cloud factor calculations used with the HRRR data. The cloud factor, used to account for atmospheric effects on the incoming solar radiation before it reaches veg and albedo corrections, was being set to 0 at night. A value of 0 for cloud factor means that it is fully cloudy, which was causing the thermal radiation to be unrealistically high, causing more melt. 

The solution was to set the cloud factor to NAN at night, and then interpolate between the daytime values using a linear method. This all happens at the station (HRRR pixel) scale after the data is read.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

All tests were performed using Docker (usdaarsnwrc/awsm:0.9.12)

- Standard unit tests that run on TravisCI
- Tested on all basins for part of WY2019 up to the first flight. 

**Test Configuration**:
* Docker 

### Summary of changes to Tuolumne

The Tuolumne was tested for WY2019 with these changes, as well as the local gradient method being applied to air_temp, vapor_pressure, and precip. The changes show a 3.7% increase in basin SWE up through 03/23/2019 before any updates were applied. This also decrease the effect of the updates from +20.7% SWE to +17.2% SWE. The temperatures of the snow were also decreased, as can be seen by an increase in the magnitude of the cold content, meaning the snowpack is less ready to melt. The changes in the location of the snow are most likely due to the addition of the local gradient method to the driving variables.

These changes are consistent with our assessment that the increased thermal radiation had been causing more melt. This change was successfull.

 tuolumne_20190323_swe_z
tuolumne_20190323_depth
tuolumne_20190323_coldcont
tuolumne_20190323_density

PR #92, opened 2019-04-18
Gridded interpolation speedup

### LINKED ISSUE #91: Local grid interpolation is slower than desired
In the San Joaquin model run for a single day, we are seeing about 25% slow down due to the calculations in the local gradient method. Profiling the `grid.py` for the unit test `test_load_data.test_grid_hrrr_local` shows that about 90% of the time spent is in the calculation of the slopes and intercepts. The new `interp_speed` changes helps to speed up the interpolation but the test case is too small to truly see the speed increases.

micah-prime:
## Description

This pull request includes the changes required to speedup the local gradient gridded interpolation method. The three main changes are:
* Extraction of the Delaunay call in scipy griddata so that it happens once per run rather than every time griddata is called.
* Reworking of the input data manipulation using Pandas to be much faster. 
* Made threading optional in the ```loadTopo.py``` when running ```viewf``` and ```gradient```. This option is now added to the config file, with default being ```True```.

The second change accounts for most, if not all of the speedup when using cubic interpolation. The first change has a larger effect when using linear interpolation. See Issue #91 for more details on timing. 

Fixes #91 

## Type of change

- [x] Bug fix (not really a bug, we wanted a faster implementation of code)

## How Has This Been Tested?

This change has been tested in 2 ways:

- [ ] Standard unit tests on Travis-CI. Recently added output variable comparisons to catch unintended changes in code. 
- [ ] Local test over 12 hours on 04/02/2019 in the Tuolumne using the HRRR data. This showed no change at all during the 12 hours for air temp, vapor pressure, and precip mass, as shown below:

gradient_update_check

**Test Configuration**:
Test A:
* Docker on Travis-CI
Test B:
* Python 3.5 on Ubuntu 16.04


scotthavens:
Should also note in the PR that the `loadTopo` was modified to provide an option to use threading when calculating gradient and viewf. As of now it is set to `False` meaning a slight slow down when SMRF initializes.

micah-prime:
Added those changes to the original PR in the first section. Should we leave threading set to False?

scotthavens:
We should probably just do it right and add it to the config in the `topo` section then update the PR.

micah-prime:
@scotthavens sounds good. I'll do that in a minute

micah-prime:
@scotthavens  That is now added to the config file under ```['topo][threading]``` with default being ```True```. Took a little longer because this upset a check in the backup of the config file. Tests are running on Travic CI now.

Issue #111, opened 2019-08-29
Requirements overhaul


scotthavens:
From `pipdeptree` here are the packages that are install from the `requirements.txt`
```
coloredlogs==6.1
  - humanfriendly [required: >=2.1, installed: 4.6]
coveralls==1.8.2
  - coverage [required: >=3.6,<5.0, installed: 4.5.4]
  - docopt [required: >=0.6.1, installed: 0.6.2]
  - requests [required: >=1.0.0, installed: 2.20.0]
    - certifi [required: >=2017.4.17, installed: 2017.11.5]
    - chardet [required: >=3.0.2,<3.1.0, installed: 3.0.4]
    - idna [required: >=2.5,<2.8, installed: 2.7]
    - urllib3 [required: >=1.21.1,<1.25, installed: 1.23]
Cython==0.28.4
flake8==3.5.0
  - mccabe [required: >=0.6.0,<0.7.0, installed: 0.6.1]
  - pycodestyle [required: >=2.0.0,<2.4.0, installed: 2.3.1]
  - pyflakes [required: >=1.5.0,<1.7.0, installed: 1.5.0]
inicheck==0.3.5
  - pandas [required: >=0.18.1, installed: 0.22.0]
    - numpy [required: >=1.9.0, installed: 1.14.0]
    - python-dateutil [required: >=2, installed: 2.6.1]
      - six [required: >=1.5, installed: 1.11.0]
    - pytz [required: >=2011k, installed: 2017.3]
mock==2.0.0
  - pbr [required: >=0.11, installed: 3.1.1]
  - six [required: >=1.9, installed: 1.11.0]
mysql-connector-python-rf==2.2.2
netCDF4==1.2.9
  - numpy [required: >=1.7, installed: 1.14.0]
pipdeptree==0.13.2
  - pip [required: >=6.0.0, installed: 19.2.3]
progressbar2==3.6.2
  - argparse [required: Any, installed: 1.1]
PyKrige==1.4.0
  - matplotlib [required: >=2.0.0, installed: 3.1.1]
    - cycler [required: >=0.10, installed: 0.10.0]
      - six [required: Any, installed: 1.11.0]
    - kiwisolver [required: >=1.0.1, installed: 1.1.0]
      - setuptools [required: Any, installed: 41.2.0]
    - numpy [required: >=1.11, installed: 1.14.0]
[...]

PR #112, opened 2019-08-29
Requirements cleanup

scotthavens:
The `requirement.txt` was a bit of a mess with a lot of unneeded packages. Issue #111 outlined the package dependencies that could be removed and the requirements were further split into the base requirements and those needed for testing and the documentation.

Along with this, the `Dockerfile` was modified to remove some packages and the `.dockerignore` added the `examples` directory which was 300+MB. With the docker optimization, the original image went from 1.86GB to 1.33GB. Further reduction can be attained by not including the `.git` folder but the testing needs that for coveralls.

All tests pass in SMRF and the documentation builds in the docker image.

This has been tested to be compatible with the [IPW pull request](https://github.com/USDA-ARS-NWRC/ipw/pull/4)


scotthavens:
This PR will also fix the security vulnerability and allow a different package to manage the version of `urllib3`

micahjohnson150:
Do you have thoughts on which version of setup tools you want to use? I ask because the newest version is deprecating python setup.py test and is asking folks to switch to tox

scotthavens:
Then probably the latest version that supports `python setup.py test` or we switch to `python -m unittest` which is a little more verbose since we are not testing different types of python versions.

micahjohnson150:
This looked good to me, who else do you want to review?

micahjohnson150:
@scotthavens should we merge this in? Or should we wait for @etrujil  to review? I just updated this with the new changes from PR #133

etrujil:
I would happily review but would need some background as for what kind of stuff I should be looking at. Just let me know. @scotthavens @micahjohnson150

.travis.yml
```diff
@@ -8,7 +8,7 @@ services:
   - docker
 
 before_install:
-  - docker build -t usdaarsnwrc/smrf .
+  - docker build --build-arg REQUIREMENTS=_dev -t usdaarsnwrc/smrf .
```
etrujil: What does this do?

Issue #114, opened 2019-09-04
Threading with gridded input data


scotthavens:
The threading function of SMRF never worked quite right with gridded data (old issue #3). While it was fixed, with the addition of HRRR inputs, it stopped working as there are no tests to ensure that it works.

scotthavens:
It appears to be a few different possible issues. First, it can get hung up in the queue with the `thermal_cloud` variable for the gridded inputs. That was an easy fix by removing the variables from the queue (not really needed at the moment).

The more complex one is in the tests. The queue doesn't appear to be properly cleaned up as the queue items like `flatwind` show up in the gridded queue. Running the gridded tests by themselves don't really have a problem and will run fine. But run the `test_configurations` then the gridded tests, it will have those `flatwind` values in the queue.

For example, `python -m unittest -v`  will not work but each test can be ran individually and will pass. This is leading me to believe that there is some cleanup that isn't happening in the tests somewhere.

scotthavens:
Closing this for now as the tests in both SMRF and AWSM are passing. Reopen if that changes after more testing.

PR #118, opened 2019-09-05
Grid threading

scotthavens:
There was a bit going on for why the threading didn't work with grid:
- the thermal cloud factors between the two different types of gridded methods
- HRRR loading loads extra data which needed to be trimmed
- The full SMRF test wasn't properly cleaning up after completion, had to add a `tearDown`
- `thread_variables` was on the `SMRF` class and needed to be moved to the instance
- Added lots of debugging logs so hopefully in the future it's easier to figure out what might be happening

Added a test which works with the WRF, HRRR and generic NetCDF files. This fixes issue #114


scotthavens:
Upon a longer run, SMRF is still hanging at random points. There has been adequate logging enabled to show that this is most likely not a queue problem. The only other thing that I can think it could be is a subprocess command that is not exiting properly but not terminating the thread. So this becomes a little bigger issue to remove some of the calls to IPW functions throughout SMRF. There are a few that are targeted at the radiation processes so it will take some time to convert them over. PR #122 is an example of converting the `sunang` calculation from C to Python.

For now, I'm closing the PR as it's not fully working.

micahjohnson150:
Dang. So you think its the IPW calls? Watch out stoporad...we coming for you...

PR #122, opened 2019-10-02
Sun angle calculation now pure Python with Numpy support

scotthavens:
The sun angle calculation was done with a subprocess call to the IPW `sunang` function. The calculations were extracted from IPW and converted to Python. Tests were added to ensure that the sun angle functions were returning correct values and improved on the functionality. If latitude and longitude are provided as a numpy array, `sunang` will calculate the sun angle for every point.

## At a point, single date
The first test replicates the example in the IPW man page for `sunang` with no difference between the example and the Python version. The values must be truncated in Python to match the print to `stdout` that SMRF reads from. The truncation can be turned off if needed or is a numpy array is passed.

## At a point, whole water year
Comparing over a full water year for RME shows very little differences, most likely in the precision of the print statements. The azimuth below shows a difference in azimuth (degrees) of less that 0.01 degrees.
smrf_sunang_azimuth

The RME zenith angle in degrees is similar, with absolute values less than 0.0002 degrees.
smrf_sunang_zenith

This text (csv) file show the difference for every hour in this test. The code used to generate the figures and text file are in `tests/test_radiation.py`.
[sunang_comparison.txt](https://github.com/USDA-ARS-NWRC/smrf/files/3682791/sunang_comparison.txt)

## Affects to net solar radiation
With the small differences, the net solar radiation is not the same as the gold standard for RME causing the tests to fail. However, looking at the differences most values are about 0 with a maximum of 0.175 W/m2 which is insignificant for the magnitude of solar radiation. Accepting this PR will require an update in the gold standard.
smrf_net_solar_diff

## Sun angle for a whole basin
With the sun angle in Python, we can now pass latitude and longitude as an array. This will return an array of `cosz` and `azimuth` that can be provided to functions when they are ready. This will require rework of illumination angle and solar modules. Below is an example from the Tuolumne River Basin. `cosz` does not vary widely across the basin but the `azimuth` does have some differences of around 1 degree from one side to the other.
smrf_sunang_cosz_tuolumne
smrf_sunang_azimuth_tuolumne

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,6 +29,7 @@ pycodestyle==2.3.1
 pyflakes==1.5.0
 Pygments==2.2.0
 python-dateutil==2.6.1
+pysolar>=0.8,<0.9
 pytz==2017.3
 pyyaml==4.2b1
 requests==2.20.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,7 +29,6 @@ pycodestyle==2.3.1
 pyflakes==1.5.0
 Pygments==2.2.0
 python-dateutil==2.6.1
-pysolar>=0.8,<0.9
 pytz==2017.3
 pyyaml==4.2b1
 requests==2.20.0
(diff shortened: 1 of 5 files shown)


scotthavens:
This PR has been pulled into PR #123

PR #125, opened 2019-10-24
Gradient calculations to python: slope and aspect

scotthavens:
In `loadTopo` the IPW `gradient` method is called to calculate the slope and aspect for use in `stoporad`. This is a fairly easy calculation that was converted over to Python. The method that IPW uses in `gradient` is a finite difference approach in just the x and y direction. Two new methods for calculating the slope and aspect are implemented.

# Gradient methods

## gradient_d4
The `gradient_d4` method mimics the IPW `gradient` performing the finite difference in only the x and y direction.

## gradient_d8
GIS frameworks like ArcGIS and GDAL use a second order finite difference, implemented in `gradient_d8`. The ArcGIS [documentation](http://desktop.arcgis.com/en/arcmap/10.3/tools/spatial-analyst-toolbox/how-slope-works.htm) has a good overview.

## Other changes
- A new configuration file option was added to the `[topo]` section for `gradient_method`, either `gradient_d4` or `gradient_d8` with the default as `gradient_d8`.
- `threading` in `[topo]` was removed
- Added tests in `test_gradient`
- The method that IPW used to pad the bottom line of the dem was a negative mirror and it made more sense to have it not be negative but continue the slope to pad the dem. Therefore, the bottom lines are different in all images and introduce some of the larger differences.

# Gradient differences
The changes in the calculation of the gradient method breaks the tests for RME and RME with HRRR. Due to the way SMRF handles the IPW `gradient` calculation with converting to IPW images and back, it was not expected for the new methods to match the old method.

## RME dem
Comparing the methods over the dem show differences between the methods. The bottom line is the most pronounced but there is minimal difference between the slop and aspect between IPW `gradient` and `gradient_d4`. There are larger differences between IPW `gradient` and `gradient_d8`. The figures below are in `tests/test_gradient.py`.

smrf_gradient_images

The historgrams show that most of the differences is close to 0 with larger spread for the `gradient_d8` method.
smrf_gradient_hist

## RME net solar radiation with gradient_d4
The gradient is used in the `stoporad` calculation which will affect the net solar radiation. The test now fail because of the differences above. The difference between the test and the net solar radiation with `gradient_d4` shows small differences between the two, most likely because of the bottom row and bit resolution of the original IPW images.

smrf_gradient_net_solar_hist_d4

## RME net solar radiation with gradient_d8
Removing the bottom row from the comparison shows that the `gradient_d8` method has a good normal like distribution centered around 0 with +/- 2 W/m2 difference.

smrf_gradient_net_solar_hist_smal

## RME HRRR net solar radiation with gradient_d8
Removing the bottom row from the comparison shows that the `gradient_d8` method has a good normal like distribution centered around 0 with +/- 5 W/m2 difference, which is a little larger than with the station data.

smrf_gradient_hrrr_net_solar_hist_small


scotthavens:
@Hedrick-ARS the tests don't pass because of the net solar. To make the tests pass we will need to update the gold files.

tests/test_gradient.py
```diff
@@ -0,0 +1,295 @@
...
+        return dem
+
+    def gen_dem_ne(self, dem_size=10):
+        dem = np.tile(range(10), (10, 1)).transpose()
+
+        for i in range(dem_size):
+            dem[i, :] = np.arange(i+dem_size, i, -1)
+        return dem
+
+    def test_gradient_d4_funcs(self):
```
jomey: Minor feedback that I have and been taught about unit testing is to break up your assertion into multiple cases, once the method gets too long. 
Has some advantages, when things start to fail. [This stack exchange](https://softwareengineering.stackexchange.com/questions/260873/how-much-should-i-break-up-my-unit-tests) names a few and shows some ideas on how to break It up.

For this instance, I would propose to have cases for each direction (north, south, ...).

scotthavens: Makes sense, when we get to reviewing this PR I'll make those changes. Not sure why I did it this way!

scotthavens: Broke these out into separate tests

jomey: AWSM ... ehh .. SMRF .. you know what I mean :-)

tests/utils/test_gradient.py
```diff
@@ -0,0 +1,333 @@
...
+        # test northeast slope and aspect
+        dem = self.gen_dem_ne()
+        py_slope, asp = gradient.gradient_d8(dem, self.dx, self.dy)
+        ipw_a = gradient.aspect_to_ipw_radians(asp)
+
+        self.assertTrue(np.all(py_slope == self.slope_val))
+        self.assertTrue(np.all(asp == 45))
+        self.assertTrue(np.all(ipw_a == (np.pi/2 + np.pi/4)))
+
+    # Used to bulid comparisons for the Pull Request
```
etrujil: What's will happen to these commented code?

scotthavens: I don't know yet. This code was used to make the plots for the PR. Maybe it would be good to start putting them into a Jupyter Notebook?

scotthavens:
> This looks like a good improvement, and a more accurate estimation of both slope and aspect.
> 
> Would it be worth adding tests with some flat or multifaceted surfaces (e.g., planes, pyramids)?

A plane would be a good test to see how it does with flat terrain. I will look into a pyramid as well, it will mainly be for testing the edges where the pyramid meets as the flat surfaces will be the same as the tests.

tests/utils/test_gradient.py
jomey: I think that is a great place to put stuff like that.
Also helps other users to get started on plots.

scotthavens:
> This looks like a good improvement, and a more accurate estimation of both slope and aspect.
> 
> Would it be worth adding tests with some flat or multifaceted surfaces (e.g., planes, pyramids)?

Added a flat slope to the gradient tests. IPW set's it convention for a flat slope to have 0 slope but south aspect.

PR #109, opened 2019-08-28
Hrrr gold standard

scotthavens:
The config file for the HRRR gold standard was not the one used to generate the standard. The config file uses the local gradient method and passes the `test_grid_hrrr_local` test. This is the fix to issue #108

Issue #108, opened 2019-08-28
HRRR gold standard unknown origin


scotthavens:
The HRRR gold standard `config.ini` does not match the dates that are in the output netcdf files. During the testing with the local gradient method, the gold standard was changed but the config file was not committed. If the tests were to compare the images, they would fail for the `test_grid_hrrr` but not for `test_grid_hrrr_local`.

PR #129, opened 2019-12-06
Hrrr gold standard

scotthavens:
Update for PR #109. The config file for the HRRR gold standard was not the one used to generate the standard. The config file uses the local gradient method and passes the test_grid_hrrr_local test. This is the fix to issue #108


scotthavens:
Closing in favor of PR #142 that includes all these changes.

PR #135, opened 2020-02-27
Improve script parameters help

jomey:
Repeating the main commit message:
```
The script argument definition used the metavar option to define short
options for parameters. This option should only be used for better help
text however and won't enable the user to use these.
This fixes the usage and also enables the shorter parameter option
names.
```


jomey:
Should have explained the intentions behind the testing, so my bad.

Main idea here is to ensure that the options defined with `ArgumentParser` are also accepted with the command line. I was stomped yesterday and doubted my ability to read command line options, when I tried to execute:
```shell
gen_maxus -W 30 ...
```
and kept getting the error that this is an unknown option.

When I checked the source code, I discovered that the `metavar` option should only be used as a label for the option in the help text and not as an additional short option.

The test ensure now that both the long format and short are accepted. So
```
gen_maxus --window
```
and
```
gen_maxus -W
```
both work. Added this to all options including checking for default values.

Hedrick-ARS:
Well then it certainly makes sense to have both options working! Thanks for changing that.

scotthavens:
And update this PR with `master` from your last PR that was merged.

Issue #142, opened 2020-03-18
Gold config update


scotthavens:
The gold config was not updated in the config file overhaul PR #126. Update the gold config file and make a script that allows for recreation of the gold files.

micahjohnson150:
Just a note like I sent in the email about the change in defaults. If the goal is only change the files that need to be changed as a result of #123 then be sure to change the defaults back for wind reduction 1.0 to 0.7 and new_snow_density from susong1999 to Marks2017 after the config file changelog is applied.

Issue #136, opened 2020-03-03
Remove changelog entry "gridded/zone_number-> removed" from changelog


scotthavens:
Trying to run a config file after applying the changes and it will not run saying that the `gridded/zone_number -> REMOVED`. The value is removed but still in the `CoreConfig` but saying it's removed in `changelog.ini`.

micahjohnson150:
So the problem here is that while it is being removed before hand, it is in the coreconfig and there is a recipe (gridded_recipe) it is being added back in. And since the changelog shows it was deleted they now conflict after applying recipes. So inicheck here is actually doing the right thing. The real issue is that I didn't make the change log agree with the core config. The question here is whether we want to adjust the changelog or remove it from the coreconfig and implement #100. Thoughts?

scotthavens:
Let's remove it from the change log for now and #100 will be the issue to remove from the core config and add to the change log.

Issue #100, opened 2019-08-26
Remove zone and Letter in favor of auto detmerination for gridded datasets config


micahjohnson150:
Pull these values from the ESPG code or it's specified in the topo section for ipw as a part of #98 
zone_number:
zone_letter:

PR #144, opened 2020-03-20
gold config updates

scotthavens:
The gold config's needed updates to the new config file sections and fields.

- changed the gold config file for the station data tests, passing the tests so the new gold files are the same as before
- updated the hrrr gold files and provided a command. Removed the zone_number and letter from the change log, fixed issue #136
- updated the gold hrrr standards with a script to generate again fixes issue #142 and fixes issue #108. This also negates PR #129 as this encompasses that PR
- The HRRR test never asserted and now it does
- scope creep, added isort and linting to the dev requirements, ran isort


robertson-mark:
97 files, my bleeding eyes! Can we do the isort and linting as a separate PR in the future?

scotthavens:
Yup, 97 files only 5 really matter so it will go quick

requirements_dev.txt
```diff
@@ -1,6 +1,8 @@
 -r requirements.txt
 coveralls>=1.3.0
-flake8==3.5.0
+flake8==3.7.9
```
robertson-mark: Minor: is it worth updating requirements_dev.txt in AWSM at the same time, which is still 3.5.0?

scotthavens: I think that is what we are trying to figure out with dependancy testing. Since these are just dev requirements and not in the final docker image, I don't see this as a big problem yet. We can do greater than 3.5.0 if that's the case

jomey: You could even think about removing the version string all together. Like you said, won't break anything

smrf/framework/changelog.ini
```diff
@@ -22,8 +22,8 @@ stations -> REMOVED
 # Gridded
 gridded/file -> gridded/wrf_file
 gridded/directory -> gridded/hrrr_directory
-gridded/zone_number -> REMOVED
-gridded/zone_letter -> REMOVED
+; gridded/zone_number -> REMOVED
+; gridded/zone_letter -> REMOVED
```
robertson-mark: Are these left commented for posterity?

requirements_dev.txt
scotthavens: Updated, there was a conflict with autopep8 but greater than should fix that

docs/conf.py
```diff
@@ -18,9 +18,12 @@
 
 import os
 import sys
-from smrf.utils.utils import get_config_doc_section_hdr
+
 from inicheck.tools import config_documentation
+# -- Have to do a mock install of some modules that RTD doesn't have --------
+from mock import Mock as MagicMock
```
jomey: Should the 'mock' lib go into the dev requirements as well?

smrf/framework/changelog.ini
scotthavens: This will be address with issue #100 so I was just leaving it in as it will go back into the change log

smrf/output/output_netcdf.py
```diff
@@ -2,17 +2,20 @@
...
+
+import netCDF4 as nc
+import numpy as np
 from spatialnc.proj import add_proj, add_proj_from_file
-# import pandas as pd
 
 from smrf import __version__
+from smrf.utils import utils
+
+# import pandas as pd
```
jomey: Nuke?

docs/conf.py
scotthavens: I didn't realize that this library is part of `unittest.mock` now. I just changed that to reflect and removed from the requirements.

smrf/output/output_netcdf.py
scotthavens: Probably, let leave it for another time when we do a good flake8 cleaning

tests/RME/gold/make_gold_stations
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jomey: Very minor and I usually follow:
```
#! /usr/bin/env bash
```

[Discussion](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash)

tests/RME/gold_hrrr/make_gold_hrrr
```diff
@@ -0,0 +1,26 @@
...
+    echo "Running SMRF"
+    run_smrf tests/RME/gold_hrrr/gold_config.ini
+
+    echo "Cleaning up"
+    rm tests/RME/gold_hrrr/log.txt
+    rm tests/RME/gold_hrrr/config.ini
+else
+    echo "RME HRRR gold files not updated"
+fi
+
```
jomey: Future iteration could combine the two bash helpers ...

tests/test_configurations.py
```diff
@@ -25,14 +30,47 @@ def can_i_run_smrf(self, config):
...
+                    getattr(test.variables[var_name], att))
+
+            # only compare those that are floats
+            if gold.variables[var_name].datatype != np.dtype('S1'):
+                result = np.abs(
+                    gold.variables[var_name][:] - test.variables[var_name][:])
+                self.assertTrue(not np.any(result > 0))
+
+        gold.close()
+        test.close()
```
jomey: Could wrap lines 44 to 61 in a `try` and `finally` block to ensure the files are closed when one of the assertions are failing.

robertson-mark:
In your initial notes you said that the tests still pass. Is it worth doing a specific comparison of the old/new gold files, similar to the extra work you had (which was awesome) in Gradient calculations to python: slope and aspect #125 for example?

tests/RME/gold/gold_config.ini
```diff
@@ -244,29 +221,22 @@ temp:                          -2.5
...
 ################################################################################
 
 [system]
 threading:                     True
 time_out:                      None
-max_values:                    2
+queue_max_values:              2
+log_level:                     debug
+log_file:                      ./log.txt
+qotw:                          False
```
robertson-mark: False?!

tests/test_data_gridded.py
```diff
@@ -0,0 +1,200 @@
...
+        output_dir = os.path.join(self.test_dir, out_dir)
+        s = os.path.join(output_dir, '*.nc')
+        file_names = glob(os.path.realpath(s))
+
+        # path to the gold standard
+        gold_path = os.path.realpath(
+            os.path.join(self.test_dir, 'RME', 'gold_hrrr'))
+
+        for file_name in file_names:
+            nc_name = file_name.split('/')[-1]
```
jomey: Same as
```
os.path.basename(path)
```
?
Basename is cross platform.

tests/RME/gold_hrrr/make_gold_hrrr
scotthavens: I think that would be a good idea as there are other gold files I'd like to add. Perhaps a makefile would be more appropriate at that point

tests/test_configurations.py
scotthavens: From my understanding, it will not raise an exception here but log that it failed. So it should always make it to the close statements. Your method would be a lot more explicit

tests/RME/gold/gold_config.ini
scotthavens: inicheck puts in the defaults!

tests/test_data_gridded.py
scotthavens: So instead of `realpath` use `basename`?

scotthavens:
@robertson-mark the logic I followed was that the tests passed for the station data before making any changes. Therefore, the tests with the new configuration files are exactly the same as gold. So if the test passes with the new gold files then the new gold files are exactly the same as the old one. Let me know if there is any reason that shouldn't be the case.

For the HRRR tests, it never compared the tests to the gold files.

tests/test_data_gridded.py
jomey: instead of `file_name.split`

tests/RME/gold/make_gold_stations
Hedrick-ARS: I like that. One of those T.I.L. moments...

PR #145, opened 2020-03-24
Timezone handling

jomey:
Went down the rabbit hole of trying to set all the timezones, when data are loaded into panda dataframes. This addressed some issues I was getting when running on CHPC, where the latest version of weather forecast retrieval is installed.

I added a few tests for `SMRF` that ensure start and end time, as well as, the `date_range` is handled as a time with zone awareness.

The tests seemed to have changed quite a bit and in the end this is the summary:
* Created a main SMRFTestCase class that is used throughout and has the the test folder and base config set as attributes.
* Broke out the MySQL check and ensure it fails faster on out of network locations.


.gitignore
```diff
@@ -101,16 +100,13 @@ docs/smrf.rst
...
+tests/RME/**/*.idx
 
 #ignore maxus.
 test_data/topo/maxus_100window.nc
 test_data/topo/maxus.nc
 examples/* *.nc
 examples/reynolds_mountain_east/output/*
 examples/simple_hill/* *.csv
-*/input_backup/*
-test_data/output/*
```
jomey: These were duplicate

smrf/data/loadGrid.py
```diff
@@ -263,7 +263,10 @@ def load_from_netcdf(self):
         # GET THE TIMES
         t = f.variables['time']
         time = nc.num2date(t[:].astype(int), t.getncattr('units'), t.getncattr('calendar'))
-        time = [tm.replace(microsecond=0) for tm in time] # drop the milliseconds
+        # Drop milliseconds and prepare to use as pandas DataFrame index
+        time = pd.DatetimeIndex(
+            [str(tm.replace(microsecond=0)) for tm in time]
+        )
```
jomey: The main change here

scotthavens: Hmmm, interesting. Better to have it more explicitly defined as a `DateTimeIndex` anyways.

jomey:
Just run into this exact same error in a run on our CHPC environment.
Could we get this in?
Alternatively I can also switch to this branch.

jomey:
Attached the [error.log.zip](https://github.com/USDA-ARS-NWRC/smrf/files/4384708/error.log.zip) that I was getting, when running on CHPC.

jomey:
Going to have another look on this. The timezone conversion errors are only happening, when I run on CHPC. Everything works on my local with same version of Python, Pandas, and the data from ERW. Only difference are the HRRR data, which are copied from the archive on CHPC and locally I use the external disk from you. Will revisit and request a review, once I have figured out what is different between the two environments.

smrf/data/loadGrid.py
```diff
@@ -297,7 +300,7 @@ def load_from_netcdf(self):
                 except:
                     pass
                 df = df[self.start_date:self.end_date]
-                setattr(self, v, df.tz_localize(tz=self.time_zone))
+                setattr(self, v, df.tz_convert(tz=self.time_zone))
```
scotthavens: You might want to try flipping these two lines. `start_date` and `end_date` are tz aware but the data frame index doesn't appear to be at this point. This is where I would debug, seeing which one of these dates is not tz aware.

jomey:
I can cherry pick the commit that updates WRF onto this branch to make tests pass again OR open a new PR with the branch linked to [the passing travis build](https://travis-ci.com/github/USDA-ARS-NWRC/smrf/builds/155715830)

scotthavens:
So this PR doesn't go way out of scope, can you make a PR for that travis build that will address issue #146?

jomey:
Opened PR #147 with only the last commit of this branch.

tests/nwrc_check.py
```diff
@@ -0,0 +1,34 @@
...
+    }
+    URL = 'http://' + MYSQL_OPTIONS['host']
+
+    @classmethod
+    def in_network(cls):
+        """
+        Checks that were on the NWRC network
+        """
+        try:
+            response = requests.get(cls.URL, timeout=1)
```
jomey: Added a 1 second timeout.
Is that long enough for the NWRC network?

scotthavens: I would hope so!

tests/smrf_test_case.py
```diff
@@ -0,0 +1,109 @@
...
+            # compare the dimensions
+            for att in gold.variables[var_name].ncattrs():
+                self.assertEqual(
+                    getattr(gold.variables[var_name], att),
+                    getattr(test.variables[var_name], att))
+
+            # only compare those that are floats
+            if gold.variables[var_name].datatype != np.dtype('S1'):
+                error_msg = f"Variable: {var_name} did not match gold standard"
+                if os.getenv('NOT_ON_GOLD_HOST') is None:
```
scotthavens: This is because of the precision error between linux and OSX?

tests/test_model_framework.py
```diff
@@ -0,0 +1,49 @@
...
+class TestModelFramework(SMRFTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.smrf = SMRF(cls.config_file)
+
+    def test_start_date(self):
+        self.assertEqual(
+            self.smrf.start_date,
+            to_datetime(self.smrf.config['time']['start_date'], utc=True)
```
scotthavens: Awesome! This still looks like it's really only testing that UTC works which was the problem before. Maybe duplicate these tests for a non UTC timezone for sanity sake?

tests/smrf_test_case.py
jomey: That is my guess as well. SciPy is throwing a precision warning and I match the first three digits compared to the gold files. I also get the error on CHPC though and that is a CentOS Linux 7.

smrf/framework/model_framework.py
```diff
@@ -246,11 +238,35 @@ def __init__(self, config, external_logger=None):
...
+
+        if not is_utz:
+            self.start_date = self.start_date.tz_localize(self.time_zone)
+            self.end_date = self.end_date.tz_localize(self.time_zone)
+
+        # Get the time steps correctly in the time zone
+        self.date_time = pd.date_range(
+            self.start_date,
+            self.end_date,
+            freq=f"{self.config['time']['time_step']}min",
```
robertson-mark, line 266: I'm unfamiliar with this, and am getting a syntax error here. Is freq=f"{60}min" what was intended?

jomey: What version of Python do you have?

jomey: To give some background - it is called an `f-string` and was introduced in Python 3.6.
https://www.python.org/dev/peps/pep-0498/#id37

robertson-mark: Makes sense, I've got 3.5.

jomey:
Mark's find makes we wonder, what versions of Python SMRF or all model components want to support?
Should formatting of Strings not use f-strings for the time being?

scotthavens:
I think that at a mimimum we should support down to 3.5 because Ubuntu 16.04 ships with 3.5 and it's still got at least another 5 years of support. While we've made the move for our containers to 18.04, there are still a lot of servers out there that are running 16.04.

This would lead to us using tox to test different versions of python, perhaps in conda since SMRF is dependent on IPW at this time.

smrf/framework/model_framework.py
```diff
@@ -263,7 +263,7 @@ def _setup_date_and_time(self):
         self.date_time = pd.date_range(
             self.start_date,
             self.end_date,
-            freq=f"{self.config['time']['time_step']}min",
+            freq="{[time][time_step]}min".format(self.config),
```
scotthavens, line 266: T.I.L. about using dictionaries in string formats!

jomey: Ditto - There are many awesome tricks in that format bag:
https://docs.python.org/3.7/library/string.html#formatstrings

PR #147, opened 2020-03-27
Docker - Install weather_forecast_retrieval via pip install.

### LINKED ISSUE #146: update weather_forecast_retrieval
`weather_forecast_retrieval` is now a pip install. This would help with the requirements and docker build.

The gold files will have to be updated as there is a small known change when going from `pygrib` to `xarray`, see the [wontfix issue](https://github.com/USDA-ARS-NWRC/weather_forecast_retrieval/issues/9). Will need to document the changes but aren't quite using the `opendap` HRRR files yet.

jomey:
The weather_forecast_retrieval (WFR) is a pip install package now and
simplifies the Docker image. Updates to WFR also requires a rebuild of
the gold images.

Fixes #146

### Changes to Gold files
Only the HRRR gold files were affected by this change. Here Histograms for each file that showed a change:
gold_hrrr_hist_air_temp nc
gold_hrrr_hist_net_solar nc
gold_hrrr_hist_precip_temp nc
gold_hrrr_hist_thermal nc
gold_hrrr_hist_vapor_pressure nc
gold_hrrr_hist_wind_direction nc
gold_hrrr_hist_wind_speed nc


jomey:
Note that there are some warnings now, when running the test suite triggered from the current version of WFR:
```
/usr/local/lib/python3.6/dist-packages/weather_forecast_retrieval/hrrr.py:716: FutureWarning: The pandas.np module is deprecated and will be removed from pandas in a future version. Import numpy directly instead
1914  data = data.assign_coords(x=pd.np.arange(0, len(data['x'])) * 3000)
```
Those will be addressed once a new version with the latest from the master branch on WFR is cut.

scotthavens:
We have been trying to document changes to the gold files so we have explicit understanding . For example, PR #125 showed the histogram for `net_solar` since that was the only variable that was changing when the gold files were updated. Can you add something similar so we all have an idea of how much the gold files changed?

scotthavens:
You can use `nc_stats` that will print out a summary difference for two files. It's in [vhdr](https://github.com/USDA-ARS-NWRC/vdhr/blob/master/scripts/nc_stats.py) at the moment.

jomey:
There is some documentation of the difference between the two different WFR version with issue #9.
Should I plot another histogram for the gold files explicitly? Any standard code checked in to ensure similar standards?

scotthavens:
> There is some documentation of the difference between the two different WFR version with issue #9.
> Should I plot another histogram for the gold files explicitly? Any standard code checked in to ensure similar standards?

Check out the comment above. In the long run this `nc_stat` will be ran when making the gold files so it's automatic.

tests/RME/plot_gold_difference_to_master.py
```diff
@@ -0,0 +1,71 @@
...
+                label=LABEL.format(
+                    difference.min(),
+                    difference.max(),
+                    difference.mean()
+                ),
+                bins=numpy.arange(-.5, .5, .001),
+            )
+            plt.title(f"{folder.replace('_', ' ').upper()}\n {file_name}")
+            plt.legend()
+            plt.savefig(f"{folder}_hist_{file_name}.png")
```
scotthavens: Awesome!

jomey: Was thinking this could be used as a hook once there is time to create a Makefile (and combine the two current bash scripts) to re-generate the gold files.

scotthavens: That's what I was thinking as well. Have a better way to generate the gold files and know how much they are changing. I like comparing to the master branch with git.

scotthavens:
The test outputs are compared to the gold files so that we know if anything changed in the way we model. For example, if we make some code improvements to solar radiation, it the tests error because they don't match the gold file, we'll know that the code improvement broke SMRF.

Issue #18, opened 2017-09-21
Migrate threaded variables to their respective modules


scotthavens:
In gitlab by @micahjohnson150 on Jul 11, 2017, 15:13

After a small issue with threaded variables after updating (see #14 ) we need to compartmentalize the threaded variables so that assigning new threaded variables for new approaches can be added with ease and avoid sprawling logic gates.

Should be moved to be assigned to the __init__ of the module so that we can have conditional creation of threaded variables.

scotthavens:
In gitlab by @micahjohnson150 on Jul 11, 2017, 15:14

removed ~1041196 label

PR #149, opened 2020-04-03
windninja topo

scotthavens:
It was discovered that Katana was flipping the topo vertically and SMRF was then flipping it back. Katana PR [#18](https://github.com/USDA-ARS-NWRC/katana/pull/18) fixed this issue in the master branch and the `release-0.3` branch.

This fix is for the `release-0.9` branch. A few things are important.

The `smrf.utils.utils.get_asc_stats` was creating the y values in ascending order. This means that the (0,0) coordinate of the numpy array (upper left) was made as the lower left in the interpolation. By reversing the y values to be descending, means that the (0,0) will be the upper left coordinate.

Additionally, ascii grids are lower left corner cell reference unlike netcdf and interpolation schemes which are cell center. While Katana accounted for that by shifting lower left coordinate by a half a cell size, SMRF did not do this. Now the output of this `get_asc_stats` will output the x and y on the same coordinates as everything else in SMRF.

After correcting this, the `np.flipud` was no longer needed as the interpolation was using the right coordinates.


smrf/distribute/wind.py
```diff
@@ -208,61 +208,67 @@ def initialize(self, topo, data):
...
+        xy = np.zeros([XW.shape[0]*XW.shape[1],2])
+        xy[:,1] = ywint
+        xy[:,0] = xwint
+        uv = np.zeros([self.X.shape[0]*self.X.shape[1],2])
+        uv[:,1] = self.Y.flatten()
+        uv[:,0] = self.X.flatten()
+
+        self.vtx, self.wts = utils.interp_weights(xy, uv,d=2)
+        ###### end do this first to speedup the interpolation later ####
+
```
Hedrick-ARS, line 271: So you moved the windninja interpolation into its own function. Nice. I like.

scotthavens: This is just the initialization of wind ninja, it used to be in the `initialize` function but this makes it easier to test

Hedrick-ARS: Ah. I see. So it just gets the coordinates and the interpolation weights.

tests/test_configurations.py
```diff
@@ -47,6 +47,42 @@ def tearDown(self):
...
+        Runs the short simulation over reynolds mountain east
+        """
+
+        cls.test_dir = os.path.join('tests', 'Lakes')
+
+        # check whether or not this is being ran as a single test or part of the suite
+        cls.config_file = os.path.join(cls.test_dir, 'config.ini')
+
+        # read in the base configuration
+        cls.base_config = get_user_config(cls.config_file, modules = 'smrf')
```
jomey, line 68: Little late to the party and I would have moved this into its own file and called the file the same as the class. Makes it easier to find the declaration in the future.
Also the description seems outdated and the commented out code below should ideally not be there.

scotthavens: This is a true patch for this release. We'll need to make these changes into the master branch and will make it look a lot prettier!

jomey: Got it. Didn't look close enough that this is a release branch and not master 👍

PR #150, opened 2020-04-05
Wind ninja interpolation

scotthavens:
There are NaN values in the resulting interpolation of the wind ninja grid to the model domain. This is because there are values that are model domain cells that are outside the domain of the wind ninja grid. The current method does not extrapolate and therefore fills 2 of the edges (left and bottom) with NaN values.

The fix here is to fill with the closest value or linearly interpolate from the closest values. Since these are always on the left and bottom edge of the image, they are not within water shed boundaries.


smrf/distribute/wind.py
```diff
@@ -590,6 +591,19 @@ def convert_wind_ninja(self, t):
         g_vel = utils.grid_interpolate(data_vel_int, self.vtx,
                                        self.wts, self.X.shape)
 
+        if np.any(np.isnan(g_vel)):
+            # https://stackoverflow.com/questions/9537543/replace-nans-in-numpy-array-with-closest-non-nan-value
+            self._logger.warning('NaN values present in the wind ninja interpolation, replaced with closest value')
+            for i in range(4):
+                mask = np.isnan(g_vel)
+                g_vel[mask] = np.interp(np.flatnonzero(mask), np.flatnonzero(~mask), g_vel[~mask])
```
etrujil, line 599: Interesting way of filling in the nan values, unfamiliar with the specifics of the method and it took me some digging to se how it would work. However, if, say, the left column is nan, it would use the second column values (closest elements) for the interpolation, but if it is filling in the last row(s) (if nans, as it's the problem being addressed at the moment), the flattened array would have nans for the last rows, and the closest values would be distant from the leftmost element of the array, and I wonder how that's actually working in the code using this.

for example:
a = [[0, 1, 2], [2, 3, 4], [np.nan, np.nan, np.nan]]
the last row would be interpolated from the y value for element in the fourth index of the flattened array (a.flat[4] that has a value of 4 in a).
- does the interpolation assign the value of a.flatten[4]  (4 in this case) to all the elements in the row? if so, is that what you intend? @scotthavens

scotthavens: It's not a perfect solution to the problem as most interpolation methods won't extrapolate. This method works the best on the left columns as it will fill in with the nearest value. Like you pointed out, the bottom rows will be a single, repeated value, in your example 4. I didn't want to make the assumption that it was always the left and bottom values just in case.

I tried a few other interpolation techniques from `scipy.interpolate` and either they wouldn't extrapolate outside the domain or there were too many inputs points from the WindNinja outputs to interpolate.

etrujil: OK, sounds good. It is good to know what would happen on different locations of the matrix

smrf/distribute/wind.py
```diff
@@ -590,6 +591,19 @@ def convert_wind_ninja(self, t):
...
+
+                if not np.any(np.isnan(g_vel)):
+                    break
```
robertson-mark, line 602: If this is intended for the outside of arrays, perhaps at this point it would be better at this point to have explicit constraints for that case?

PR #151, opened 2020-04-06
Silence IntegrationWarning from scipy.

jomey:
The radiation method can raise an IntegrationWarning, when roundoff errors are detected during the calculation. These are different depending on the platform/OS the code is running. This silences this warning during run time and logs it at debug level.

### DIFF
--- a/smrf/envphys/radiation.py
+++ b/smrf/envphys/radiation.py
@@ -7,12 +7,13 @@ import datetime
 import logging
 import math
 import os
+import warnings
 import subprocess as sp
 
 import numpy as np
 import pandas as pd
 import pytz
-from scipy.integrate import quad
+from scipy.integrate import quad, IntegrationWarning
 from scipy.interpolate import Akima1DInterpolator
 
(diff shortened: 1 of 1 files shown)


smrf/envphys/radiation.py
```diff
@@ -1070,8 +1071,16 @@ def solint(a, b):
...
-    intgrl, ierror = quad(c, a, b, limit=120)
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter('always', category=IntegrationWarning)
+        # Take the integral between the two wavelengths
+        intgrl, ierror = quad(c, a, b, limit=120)
+
+        if len(w):
+            log = logging.getLogger(__name__)
+            log.debug(
+                "Integration warning raised in solar radiation calculation"
```
scotthavens, line 1082: Maybe "Integration round off precision warning" or something a little more verbose as to the warning. If I were to just see this warning without knowing what it was, I would think that there was a problem with the integration.

Also, switch to the warning logger since it's more in line with the message.

Issue #97, opened 2019-08-22
Remove Basin Lat Long in favor of on the fly calculation


micahjohnson150:
We would like to remove the basin lat and long from the config file in exchange for calculating it on the fly from the topo.nc

micahjohnson150:
@scotthavens how do you feel about this being abstracted and added to spatialnc? 
It feels to me like it might be used in other places besides smrf, at the very least it might be used by you guys for papers and analysis.

scotthavens:
Where else does this get used? I don't see this being used many other places at a the moment. I can see #100 being moved to spatialnc since katana and others would use it.

micahjohnson150:
The function I wrote to calculate the center is what I was thinking might go into spatialnc. You are right I don't think it gets used anywhere but I thought if might be useful for general analysis stuff. 

Thats a good call on #100, I will try to add that to spatialnc.

scotthavens:
The basin lat/lon is only used in the solar angle calucation that was improved on it PR #141. Down the road, we can calculate the sun angle for every point on the basin which would allow us to remove this.

PR #154, opened 2020-04-15
Auto assign basin projection attributes and basin center

micahjohnson150:
In #100 and #97  we are looking to remove config file options in favor of obtaining them from either the topo.nc file or from auto calculation specifically relating to the basin's center (lat lon) and the projection attributes (zone number letter etc).

I added a simple test confirming the lat lon calculations as well to test_load_data.py 

This will impact the net_radiation gold file as the new lat lon is slightly different. So review the code and then I will do the analysis showing the changes to the gold files.


micahjohnson150:
The builds wont pass here until we adjust the gold files. I confirmed the failures are due to net solar in the local_hrrr and the normal RME run.

requirements.txt
```diff
@@ -7,7 +7,7 @@ numpy>=1.14.0
 pandas>=0.22.0
 pytz==2017.3
 scipy>=1.0.0
-utm==0.4.0
+utm==0.5.0
```
scotthavens, line 10: Is there something in 0.5.0 that is needed or does 0.4.0 still work?

smrf/data/loadGrid.py
```diff
@@ -63,18 +59,16 @@ def __init__(self, dataConfig, topo, start_date, end_date,
...
-        # get the zone number and the bounding box
-        u = utm.from_latlon(topo.topoConfig['basin_lat'],
-                            topo.topoConfig['basin_lon'],
-                            self.force_zone_number)
-        self.zone_number = u[2]
-        self.zone_letter = u[3]
+        self.lat = topo.basin_lat
+        self.lon = topo.basin_long
+        self.zone_number = topo.zone_number
+        self.northern = topo.northern_hemisphere
```
scotthavens, line 66: Can we make all of these variables on `self` more verbose? Or are they even needed? If we put `topo` on `self` can we just call `self.topo.basin_lat`? Then it's more verbose and we're not copying variables.

smrf/utils/utils.py
```diff
@@ -464,3 +464,36 @@ def grid_interpolate_deconstructed(tri, values, grid_points, method='linear'):
         return CloughTocher2DInterpolator(tri, values)(grid_points)
     elif method == 'linear':
         return LinearNDInterpolator(tri, values)(grid_points)
+
+def get_center(ds, mask_name=None):
```
scotthavens, line 468: Will this be used anywhere else in SMRF? I see this a very specific to the `topo` class and it's probably better suited to live as a method in the `topo` class.

tests/test_load_data.py
```diff
@@ -307,6 +310,31 @@ def test_grid_netcdf(self):
 
         self.assertIsNone(run_smrf(config))
 
+class TestLoadTopo(unittest.TestCase):
```
scotthavens, line 313: Let's move this to it's own test file like `test_topo.py`.

tests/test_load_data.py
```diff
@@ -307,6 +310,31 @@ def test_grid_netcdf(self):
...
+            'northern_hemisphere':True,
+        }
+
+        self.topo = loadTopo.topo(
+            topo_config,
+            calcInput=False,
+            tempDir= join(self.test_dir, 'RME/output')
+        )
+
+    def test_auto_calc_lat_lon(self):
```
scotthavens, line 329: This appears to just be testing if the topo.nc has a mask, lets add a test for when we don't have the mask or turning off that mask to get the center of the domain.

tests/test_load_data.py
```diff
@@ -307,6 +310,31 @@ def test_grid_netcdf(self):
...
+        '''
+        Test we calculate the basin lat long correctly
+        '''
+        # Original RME
+        # basin_lon:                     -116.7547
+        # basin_lat:                     43.067
+        self.assertTrue(self.topo.basin_lat == 43.06475372378507)
```
scotthavens, line 336: What I'm finding with floating point comparisons is that it doesn't always come out the same with different operating systems. There is [assertAlmostEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual) in the `unittest` but also in numpy there is a similar version.

micahjohnson150: I had that in there but I remembered that we were going for more exact so I went back. I am fine with it, how many decimal places do you want?

jomey: When I run other tests that compares decimals on my Mac, I was able to match the first three. Speaking for the OS X users out there :-)

scotthavens: I was thinking of you Joe! Lets try with the standard comparison with 7 decimals until it becomes too restrictive. Joe's was for the gold files which have a lot of calculations on them.

jomey: I agree and start bigger here. Lat/Lon calculations hopefully have more agreement on decimals (Mac's would otherwise navigate to different places with low decimal matching)

smrf/data/loadGrid.py
micahjohnson150: They are needed. Would you rather just assign `self.topo = topo` and then use it that way in all the place its used? e.g. 

```
 ur = np.array(utm.to_latlon(np.max(self.x), np.max(self.y), self.zone_number, northern=self.northern))
```
would become 

```
 ur = np.array(utm.to_latlon(np.max(self.topo.x), np.max(self.topo.y), self.topo..zone_number, northern=self.topo.northern_hemisphere))
```

When I first wrote it I was aiming to reduce the number of lines modified. So this will have a slightly farther reach in the file.

scotthavens: This is the way we use it elsewhere as well, it makes it cleaner and more consistent

smrf/utils/utils.py
micahjohnson150: I mentioned this a bit in the issue but originally I was thinking it would be useful for other analysis. But I can move it in.

smrf/data/loadGrid.py
micahjohnson150: Sounds good.

smrf/utils/utils.py
scotthavens: If we need for something else down the road then we can move it. I'm starting to lean towards coding for now and not something that doesn't exist quite yet, but can be convinced otherwise

smrf/data/loadGrid.py
jomey: Ultimately all the topo related info could be moved into it's own class and have all the related calculations done there. Makes easier testing as well.
This is long term wish list thinking ....

smrf/utils/utils.py
micahjohnson150: I should have elaborated more. In the issue we concluded to not move it out for this same reason. At the time I had moved it into the utils as I was thinking of moving it to spatialnc since we decided not to I just left it. But as you suggested the correct conclusion is to move all the way into the topo class. I agree, preferring current code over ephemeral-might-be code is a good strategy. lol

smrf/data/loadGrid.py
```diff
@@ -97,12 +87,15 @@ def model_domain_grid(self):
...
-                                         self.dataConfig['zone_letter'])
+        dlat[0], dlon[0] = utm.to_latlon(np.min(self.topo.x),
+                                         np.min(self.topo.y),
+                                         self.topo.zone_number,
+                                         northern=self.topo.northern_hemisphere)
+
+        dlat[1], dlon[1] = utm.to_latlon(np.max(self.topo.x),
+                                         np.max(self.topo.y),
+                                         self.topo.zone_number,
+                                         northern=self.topo.northern_hemisphere)
```
jomey, line 98: Correct me if I am wrong here, but the variable value assignments look identical to the one in the `init` call.
Can this be combined or variables re-used from the `init`?

micahjohnson150: Looks like it. That was there before hand so I was mostly on a ctrl+f and replace mission. But I think it worth DRY-ing out while I am here.

micahjohnson150:
Ok I finally have the gold analysis. I took a little side journey to add to Joes gold diff script. I just always assumed you couldn't switch branches like that while running your code. It turns out to be a super useful for thing for me in a quite a few other areas. So I have added it all to a package called goldmeister in our group. 


As for my changes here you go.


The new calculation for the basin lat and long is slightly different than what we see in the config file. So there will be net solar changes.  Below are my changes to the gold files as the gold config files for hrrr and station data are committed currently. 

**projections_updates >> tests/RME/gold/gold_config.ini**

net_solar nc_net_solar


**projections_updates >> tests/RME/gold_hrrr/gold_config.ini**

Note, included here are the other changes reported in addition to net_solar but I believe these are the discrepancies we have seen from computer to computer for HRRR data.

net_solar nc_net_solar
precip_temp nc_precip_temp
thermal nc_thermal
vapor_pressure nc_vapor_pressure




I wanted to be sure in my changes so I went back to the master branch and changed the config file to the lat long that I am calculating to see the changes were identical. 

**master >> tests/RME/gold/gold_config.ini [MODIFIED]**

net_solar nc_net_solar

**master >> tests/RME/gold_hrrr/gold_config.ini [MODIFIED]**
net_solar nc_net_solar
precip_temp nc_precip_temp
thermal nc_thermal
vapor_pressure nc_vapor_pressure

scotthavens:
It appears that changes to `precip_temp`, `thermal` and `vapor_pressure` are just float point precision?

micahjohnson150:
@scotthavens yep, that was my interpretation. The hrrr tests when not on the host machine that made them only tests to 0.003 precision. I just didn't know what the protocol is for this situation. I almost left them out but I wanted to show them incase someone disagreed with them being float precision.

micahjohnson150:
Actually now that I say this, I am assuming you made the gold images for hrrr and you machine is the host? If thats the case and we want to keep that, we will want you to commit the gold file changes when we decide to do so.

scotthavens:
@jomey was the last one to change the gold files I believe but this is just a problem that we will have to figure out.

To take host machines out of it, maybe we build the gold files with the docker image?

jomey:
> @jomey was the last one to change the gold files I believe but this is just a problem that we will have to figure out.
> 
> To take host machines out of it, maybe we build the gold files with the docker image?

I used the docker image to rebuild the gold files.

micahjohnson150:
Ok I am doing better now. Fewer tiny changes. Thanks for the tip. The following results are comparison from my branch projections update to master. The data was remade using the docker. 

Here it is for the station gold: 
net_solar nc_net_solar
 
Here it is for the HRRR data
net_solar nc_net_solar
thermal nc_thermal

scotthavens:
Looks a lot better with just `net_solar`. A little odd that `thermal` is different but it's all in the noise.

tests/test_load_data.py
```diff
@@ -146,7 +149,6 @@ def test_grid_wrf(self):
 
         adj_config = {'gridded': {'data_type': 'wrf',
                                   'wrf_file': './RME/gridded/WRF_test.nc',
-                                  'zone_number': '11',
                                   'zone_letter': 'N'},
```
robertson-mark, line 152: Can zone_letter come out too?

PR #160, opened 2020-06-05
Topocalc for sky view factor and shade

scotthavens:
Many of the topo calculations were pulled out of SMRF and moved into [topocalc](https://github.com/USDA-ARS-NWRC/topocalc). The functions implemented in `topocalc` are:
- gradient
- sky view factor
- shade

These functions have been removed from SMRF.

Big changes are:
- sky view factor is calculated with more angles (72 instead of 16) and has a significant difference on the net solar and thermal.
- All other variables did not change
- `gold_difference` command line to compare gold files
- clean up of `loadTopo` and removal of `calcInputs` that were used to bypass the sky view factor calculation. Now the calculation is quick.
- `flake8` for files I was working on
- cleaned up the extremely messy `radiation.py` and broke it into a few smaller files

# RME gold changes

Net solar and thermal changed not that drastically due to the small size and relatively flat basin.

RME_gold_net_solar

RME_gold_thermal

# RME HRRR

Net solar and thermal changed not that drastically due to the small size and relatively flat basin.

RME_gold_hrrr_net_solar

RME_gold_hrrr_thermal

# Lakes HRRR

Lakes is a much larger basin than RME with many steep areas. The net solar changed quite dramatically. Steeper northern slopes see an increase in solar radiation and southern slopes see a decrease in solar radiation. With thermal radiation, most of the values are close to 0 with localized areas of decreased thermal radiation.

Lakes_gold_hrrr_net_solar

Lakes_gold_hrrr_thermal


examples/SMRF_dewPoint.py
```diff
@@ -16,28 +16,28 @@ def main():
...
-            #===============================================================================
+            # ===============================================================================
 
             #
             # These are steps that will load the necessary data and initialize the framework
             # Once loaded, this shouldn't need to be re-ran except if something major changes
 
             # load topo data
-            s.loadTopo(calcInput=False)
+            s.loadTopo(=False)
```
etrujil, line 34: what's the reason to delete the param?

scotthavens: A mistake! I should have deleted the whole thing.

scotthavens: All fixed.

jomey:
Quick question what the general idea for the `examples` folder is? Mainly asking to get a better idea of what should be in there and the audience for that.

scotthavens:
The examples folder was mean long ago to show users how to use the API functionality of SMRF to run single variables. SMRF is heavily focused on running all the variables so the API useage is minimal. I'm wondering if these examples would be better served in the documentation rather than the repo because it is unsure (and highly likely) that these do not work.

tests/smrf_test_case.py
```diff
@@ -117,6 +117,11 @@ def setUpClass(cls):
         # read in the base configuration
         cls.base_config = get_user_config(config_file, modules='smrf')
 
+        # create the output dir
+        folder = os.path.join(cls.base_config.cfg['output']['out_location'])
+        if not os.path.exists(folder):
+            os.mkdir(folder)
```
jomey, line 123: There is a call within Python >= 3.2 to make this one line:
https://docs.python.org/3/library/os.html#os.makedirs
```python
os.makedirs(folder, exist_ok=True)
```

jomey:
> The examples folder was mean long ago to show users how to use the API functionality of SMRF to run single variables. SMRF is heavily focused on running all the variables so the API useage is minimal. I'm wondering if these examples would be better served in the documentation rather than the repo because it is unsure (and highly likely) that these do not work.

Might be a scope creep for this one and worth putting on a ticket to (re)move the examples. Less code to maintain and review ;-)

setup.py
```diff
@@ -151,6 +151,7 @@ def c_name_from_path(location, name):
         'scripts/update_configs',
         'scripts/run_smrf',
         'scripts/mk_project',
-        'scripts/gen_maxus'
+        'scripts/gen_maxus',
+        'scripts/gold_difference'
```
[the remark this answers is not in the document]
scotthavens: This is true. We probably don't need users changing gold files

smrf/data/loadTopo.py
```diff
@@ -88,24 +66,7 @@ def readNetCDF(self):
...
-                v_file = v_smrf
-
-            if v_file in f.variables.keys():
-                if v_smrf == 'veg_type':
-                    result = f.variables[v_file][:].astype(int)
-                else:
-                    result = f.variables[v_file][:].astype(np.float64)
-
-            setattr(self, v_smrf, result)
+        self.readImages(f)
```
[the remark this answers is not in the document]
scotthavens: Didn't think that the method name was explanation enough... There are a lot of these leftover in all our code so it'll take some time to work though these so beat away

smrf/data/loadTopo.py
```diff
@@ -172,147 +162,122 @@ def get_center(self, ds, mask_name=None):
...
+            - sky view factor
+            - terrain factor
         """
 
-        func = self.topoConfig['gradient_method']
-        dx = np.mean(np.diff(self.x))
-        dy = np.mean(np.diff(self.x))
+        # DEM ipw image
+        dem_file = os.path.join(self.tempDir, 'dem.ipw')
+        i = ipw.IPW()
```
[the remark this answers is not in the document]
scotthavens: I thought about changing that, then didn't, then did it the right way

smrf/envphys/radiation_ipw.py
```diff
@@ -0,0 +1,168 @@
...
+import subprocess as sp
+
+import numpy as np
+on_rtd = os.environ.get('READTHEDOCS') == 'True'
+if on_rtd:
+    IPW = '.'  # placehold while building the docs
+elif 'IPW' not in os.environ:
+    IPW = '/usr/local/bin'
+else:
+    IPW = os.environ['IPW']     # IPW executables
```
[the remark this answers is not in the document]
scotthavens: It could probably be done in setup.py? The IPW executable aren't always on the path so it makes it tough and subprocess needs to know that env var. There are now only 2 calls after this PR to IPW, so this could be removed at some point.

tests/smrf_test_case.py
scotthavens: Awesome!

PR #164, opened 2020-06-11
Time zone capitilization

scotthavens:
`inicheck` parses the `time_zone` value to lowercase, which is fine for UTC but not any other time zones. Added a custom parser that will perform no checking and return the raw string provided. `pytz` will error if the timezone cannot be found.


tests/test_model_framework.py
```diff
@@ -53,15 +53,21 @@ class TestModelFrameworkMST(SMRFTestCase):
     """
     Test timezone handling for MST.
     """
-    TIMEZONE = pytz.timezone('US/Mountain')
```
jomey, line 56: Little did I know that MST is the way to go and 'US/Mountain' is deprecated

scotthavens: `US/Mountain` will also apply daylight savings time so you can either miss an hour or double up.

tests/test_model_framework.py
```diff
@@ -53,15 +53,21 @@ class TestModelFrameworkMST(SMRFTestCase):
...
+    TIMEZONE = pytz.timezone('MST')
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         base_config = copy.deepcopy(cls.base_config)
-        base_config.cfg['time']['time_zone'] = str(cls.TIMEZONE)
+        base_config.cfg['time']['time_zone'] = 'MST'
```
jomey, line 62: Did this call need to b changed?
Or did you want to create a more 'real world' input scenario with it?

smrf/utils/utils.py
```diff
@@ -39,15 +33,34 @@ def type_func(self, value):
...
+    def type_func(self, value):
+        """
+        Do not change the passed value at all
+        Args:
+            value: A single string
+        Returns:
+            value: A single string unchanged
+        """
+
+        return value
```
jomey, line 61: Think this would be worth opening a ticket for and move into 'inicheck'

tests/test_model_framework.py
scotthavens: It's a more real world application as we don't typically pass a pytz object. This will also test that the inicheck parser is working. Then `TIMEZONE` becomes what is checked against

smrf/utils/utils.py
scotthavens: Agreed.

https://github.com/USDA-ARS-NWRC/inicheck/issues/55

jomey: USDA-ARS-NWRC/inicheck/issues/54

jomey: Now we can flip a coin :-)

scotthavens: You got a lower number so you won!

PR #165, opened 2020-06-17
stoporad to Python

scotthavens:
SMRF uses `stoporad` from IPW to estimate the clear sky radiation over snow. This PR implements a full Python version of 3 main IPW utilities: `elevrad`, `toporad` and `stoporad`. 

# TL;DR

- `elevrad` is within the bit resolution noise of the IPW `elevrad`
- `toporad` is within the bit resolution noise of the IPW `toporad`
- `stoporad` has differences of +/- a few `W/m2` due to slightly different calculations for `horizon` and `albedo`
- New file structure in `envphys` to split up the monster `radiation.py` into manageable pieces that are also more intuiative (see below)
- Fixed an issue with the `loadTopo`, the new `basin_setup` has the `y` direction decreasing which creates a negative `dy`. This negative `dy` is passed to the gradient function which flips everything up/down.
- The threaded aspect of `distribute.solar` was simplified as seperate threads for the IPW calls are no longer needed.
- The Python version is 20x faster than the IPW version

## New file structure in `envphys`

Takes the old `radiation.py` and seperates into a `solar` and `thermal` package with organized files.

```
smrf/envphys/
├── albedo.py
├── constants.py
├── core
│   ├── dewpt.c
│   ├── envphys_c.c
│   ├── envphys_c.h
│   ├── envphys_c.pyx
│   ├── envphys.h
│   ├── iwbt.c
│   └── topotherm.c
├── precip.py
├── snow.py
├── solar
│   ├── cloud.py
│   ├── ipw.py
│   ├── irradiance.py
│   ├── model.py
│   ├── toporad.py
│   ├── twostream.py
│   └── vegetation.py
├── storms.py
├── sunang.py
├── thermal
│   ├── clear_sky.py
│   ├── cloud.py
│   ├── topotherm.py
│   └── vegetation.py
└── vapor_pressure.py
```

# stoporad validation

Validation of the new functions were performed against the IPW versions. Everything was held constant between the two as much as possible. A notebook containing the comparisons are in `notebooks/toporad_comparison.ipynb`. All comparisons were done in the Lakes. 

## elevrad

`elevrad` is the spatial version of `twostream` which was implemented in PR #123. Comparison with the IPW version show bit resolution noise from the 8-bit IPW output image. All values are in `W/m2`.

image

## toporad

`toporad` uses the output of `elevrad` and adjusts for topography affects. Comparison with the IPW version show bit resolution noise from the 8-bit IPW output image. All values are in `W/m2`.

image

## stoporad

`stoporad` in IPW is a [shell script](https://github.com/USDA-ARS-NWRC/ipw/blob/master/src/bin/rad.bas/stoporad/stoporad.sh) that runs multiple functions, including `elevrad` and `toporad`. The `stoporad` function calls are specific for either the visible or infrared wavelengths. There are differences in both the visible and infrared images but are minimal. These differences are caused by using a different `horizon` and `albedo` function before calling `elevrad` and `toporad`. For the visible, the largest differences are single pixels that are different from the two `horizon` outputs. All values are in `W/m2`.

image

image

# Gold changes

## RME stations
All variables are 0 expect net solar with minimal changes.

net_solar nc_net_solar

## RME HRRR
All variables are 0 expect net solar with minimal changes.

net_solar nc_net_solar

## Lakes HRRR
All variables are 0 except net solar and thermal. The cause for the large changes were an issue with `loadTopo` as the `topo.y` variable is decreasing, hence dy is negative which is passed to the gradient calculation. The RME y increases so this wasn't noticed.

net_solar nc_net_solar

thermal nc_thermal


smrf/distribute/solar.py
```diff
@@ -268,38 +278,36 @@ def distribute(self, t, cloud_factor, illum_ang, cosz, azimuth,
...
 
-            val_beam, val_diffuse = self.calc_vis(min_storm_day, wy_day,
-                                                  tz_min_west, wyear, cosz,
-                                                  azimuth)
+            val_beam, val_diffuse = self.calc_stoporad(
+                date_time, illum_ang, cosz, azimuth,
+                albedo_vis, self.VISIBLE_WAVELENGTHS)
+
             setattr(self, 'clear_vis_beam', val_beam)
             setattr(self, 'clear_vis_diffuse', val_diffuse)
```
jomey, line 312: This is mainly observational, and I like to use `@property` decorator in classes and then have a more readable code:
https://pythonguide.readthedocs.io/en/latest/python/property.html#property

smrf/distribute/solar.py
```diff
@@ -192,11 +186,35 @@ class solar(image_data.image_data):
...
+        'veg_vis_beam',
+        'veg_vis_diffuse'
+    ]
+
+    CLOUD_THREAD_VARIABLES = [
+        'cloud_ir_beam',
+        'cloud_ir_diffuse',
+        'cloud_vis_beam',
+        'cloud_vis_diffuse'
+    ]
```
jomey, line 211: Sweet!

smrf/distribute/solar.py
```diff
@@ -584,10 +494,44 @@ def calc_net(self, albedo_vis, albedo_ir):
         self.net_solar = vv_n + vir_n
         self.net_solar = utils.set_min_max(self.net_solar, self.min, self.max)
 
-    def calc_ir(self, min_storm_day, wy_day, tz_min_west, wyear, cosz, azimuth):
+    def calc_stoporad(self, date_time, illum_ang, cosz, azimuth,
+                      albedo_surface, wavelength_range=[0.28, 0.7]):
```
jomey, line 498: Can't you use the constant from above here?
If not do something like
```python
def calc_stoporad(self, ...., wavelength_range=None)

if wavelength_range is None:
    wavelength_range = self. VISIBLE_WAVELENGTHS
```

smrf/envphys/constants.py
```diff
@@ -0,0 +1,20 @@
+# define some constants
```
jomey, line 1: Really :-)

smrf/envphys/solar/cloud.py
```diff
@@ -0,0 +1,150 @@
...
+    # computationally expensive
+    # cs_solar = df_solar.copy()
+    # for dt, row in cs_solar.iterrows():
+    #     dtt = pd.to_datetime(dt)
+    #     for ix,value in row.iteritems():
+    #         # get solar using twostream only if the sun is
+    # up
+    #         if value > 0:
+    #             cs_solar.loc[dt, ix] = model_solar(dtt,
+    # df_meta.loc[ix, 'latitude'], df_meta.loc[ix, 'longitude'])
```
jomey, line 89: Any reason to keep the commented code around? (below too)

smrf/envphys/solar/irradiance.py
```diff
@@ -0,0 +1,256 @@
+import logging
+import warnings
+
+import numpy as np
+from scipy.integrate import IntegrationWarning, quad
+from scipy.interpolate import Akima1DInterpolator
+
+from smrf.envphys import sunang
+
+SOLAR_CONSTANT = 1368.0  # solar constant in W/m**2
```
jomey, line 10: Move to `constants.py`?

smrf/envphys/solar/toporad.py
```diff
@@ -0,0 +1,285 @@
...
+from smrf.envphys.thermal.topotherm import hysat
+from smrf.envphys.albedo import albedo
+from smrf.envphys.constants import SEA_LEVEL, STD_LAPSE, \
+    GRAVITY, MOL_AIR, STD_AIRTMP
+
+VISIBLE_MIN = .28
+VISIBLE_MAX = .7
+IR_MIN = .7
+IR_MAX = 2.8
+
```
jomey, line 16: Same constants have been defined in `smrf/distribute/solar.py`
Combine?

jomey: Or move to `constants.py`?

smrf/envphys/solar/toporad.py
```diff
@@ -0,0 +1,285 @@
...
+def stoporad_ipw(date_time, tau_elevation, tau, omega, scattering_factor,
+                 wavelength_range, start, current_day, time_zone,
+                 year, cosz, azimuth, grain_size, max_grain, dirt, topo):
+    """stoporad simulates topographic radiation over snow-covered terrain.
+    Uses a two-stream atmospheric radiation model. This function mimics the
+    original IPW stoporad program and differences are bit noise resolution
+    of the 8-bit IPW images.
+
+    TODO: depricated, mainly for testing
```
jomey, line 89: Typo: deprecated

(Spelling found in more places)

smrf/envphys/thermal/topotherm.py
```diff
@@ -0,0 +1,117 @@
...
+    20151027 Scott Havens
+    """
+
+    t_prime = ta - (l * z)
+    rh = ea / sati(ta)
+    rh[rh > 1] = 1
+
+    e_prime = (rh * sati(t_prime))/100.0
+
+    air_emiss = (1.24*np.power(e_prime/t_prime, 1./7.0))*pa/SEA_LEVEL
```
jomey, line 31: Could use some linting ....

smrf/framework/model_framework.py
```diff
@@ -782,17 +770,17 @@ def create_distributed_threads(self):
...
+        # t.append(Thread(
+        #     target=self.distribute['solar'].distribute_thread_clear,
+        #     name='clear_vis',
+        #     args=(q, self.data.cloud_factor, 'clear_vis')))
+
+        # # 7.2 Clear sky ir
+        # t.append(Thread(
+        #     target=self.distribute['solar'].distribute_thread_clear,
+        #     name='clear_ir',
+        #     args=(q, self.data.cloud_factor, 'clear_ir')))
```
jomey, line 783: Nuke ... 💣

smrf/utils/pycompat.py
```diff
@@ -36,10 +36,9 @@ def itervalues(d):
...
 try:
-    from cyordereddict import OrderedDict
-except ImportError:  # pragma: no cover
-    try:
-        from collections import OrderedDict
-    except ImportError:
-        from ordereddict import OrderedDict
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
```
jomey, line 44: Is there still Python 2 support in SMRF?

tests/envphys/solar/test_toporad.py
```diff
@@ -0,0 +1,226 @@
...
+            cls.cosz)
+
+    def setup(self):
+        super().setup()
+
+        self.assertTrue(self.tau_elevation == 100.0)
+        self.assertTrue(self.tau == 0.2)
+        self.assertTrue(self.omega == 0.85)
+        self.assertTrue(self.scattering_factor == 0.3)
+        self.assertTrue(self.surface_albedo == 0.5)
```
jomey, line 68: Why is this tested with each call to `test_XXX`?

scotthavens:
> Wondering if there is an opportunity to use much of the constants or standard solar calculations from the astropy package?
> https://www.astropy.org

Perhaps. At one time I did look around at other packages but they seem to either be tailored to astronomy or solar panel applications. I currently haven't found a package that is specific the the problem that the the `toporad` functions solve. However, the constants should almost all be the same but should we have another dependency just for constants?

jomey:
However, the constants should almost all be the same but should we have another dependency just for constants?

I agree that a dependency for constants is not worthwhile. Mostly wondering more about the standard solar angle calculations , since those seem standard.

scotthavens:
I think it would be something to look at in the future and see how they compare. Since you'll be working with albedo which is a calculation dependent on the solar angles, might be worth looking into it.

jomey:
> I think it would be something to look at in the future and see how they compare. Since you'll be working with albedo which is a calculation dependent on the solar angles, might be worth looking into it.

#166

requirements_dev.txt
```diff
@@ -8,6 +8,6 @@ pydata-sphinx-theme
 sphinxcontrib-bibtex>=1.0
 sphinxcontrib-websupport>=1.0.1
 pygit2
-jupyter
+jupyterlab
```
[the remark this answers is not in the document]
scotthavens: Cool, I didn't know you could add virtualenv's to Jupyter. That makes a lot more sense that this route. Removing the dependency and also seaborn.

smrf/distribute/solar.py
```diff
@@ -268,38 +278,36 @@ def distribute(self, t, cloud_factor, illum_ang, cosz, azimuth,
...
-                :mod:`smrf.distribute.precip.ppt.last_storm_day_basin`
             albedo_vis: numpy array for visible albedo, from
                 :mod:`smrf.distribute.albedo.Albedo.albedo_vis`
             albedo_ir: numpy array for infrared albedo, from
                 :mod:`smrf.distribute.albedo.Albedo.albedo_ir`
 
         """
 
-        self._logger.debug('%s Distributing solar' % t)
+        self._logger.debug(f'{date_time} Distributing solar')
```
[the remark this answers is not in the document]
scotthavens: Dang, probably shouldn't. Didn't realize it was a 3.6 thing.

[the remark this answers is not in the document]
jomey: I also wish this would have been added sooner: https://www.python.org/dev/peps/pep-0498/

smrf/distribute/solar.py
```diff
@@ -6,7 +6,7 @@
 from spatialnc import ipw
 
 from smrf.distribute import image_data
-from smrf.envphys import radiation
+from smrf.envphys.solar import cloud, vegetation, toporad
 from smrf.utils import utils
```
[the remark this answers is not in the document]
scotthavens: scope has been crept.

smrf/distribute/solar.py
scotthavens: So `calc_stoporad` would be a property? I'm not all that familiar with the `@property`

jomey: yup, the `@decorator` is basically a wrapper for the above call.

Would look something like this:
```python
class Solar():
    def __init__(self, ...):
        self. calc_stoporad = value_from_init # Can also be `None` if not given with init

    @propert
    def calc_stoporad(self):
        return self. _calc_stoporad

    @ calc_stoporad.setter
    def calc_stoporad(self, val):
        self._radius = val
```

smrf/distribute/solar.py
scotthavens: Added the two wavelength constants to the `constant` file just to keep everything grouped together.

smrf/envphys/solar/cloud.py
scotthavens: Removed, not needed anymore. However, it might be good to implement another cloud factor routine we have #167 that isn't in any of our repos.

smrf/envphys/solar/irradiance.py
scotthavens: Moved.

smrf/envphys/solar/toporad.py
scotthavens: Moved to constants.

smrf/envphys/solar/toporad.py
scotthavens: Dang spell checker didn't catch this...

requirements_dev.txt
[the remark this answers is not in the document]
Hedrick-ARS: Really awesome to see these large packages getting the boot. Thanks @jomey for going through all this and giving suggestions

smrf/envphys/thermal/topotherm.py
scotthavens: some linting and more verbose variable names

smrf/utils/pycompat.py
scotthavens: No, but there are some leftovers from when there was. I only did this since VScode debugger was reporting an error every time. This should be cleaned up to remove the compatibility.

tests/envphys/solar/test_toporad.py
scotthavens: The tests were passing individually but not when the whole class was ran so I was making sure things weren't changing. Turned out that a .copy() was needed in stoporad. I've removed.

scotthavens:
> Really nice improvement to the readability of the radiation component!
> 
> One thought I want to throw out there is to add a ticket to improve performance by vectorizing a lot of those functions? Without looking much into it, most calculations look like they can be easily converted with numba for instance.

I've tried to keep them as vectorized as possible to take advantage of numpy. Numba could definitely help on the larger basins or high resolution DEM's perhaps. As it stands, the straight python versions of these functions are 20x faster than the IPW versions. A 4 month simulations I did for a 500x1000 domain completed in like an hour instead of 3-4 hours (and not even threaded). So we're already seeing huge speed ups from this.

The tests freaked me out a little bit because the tests went from ~90 seconds to ~13 seconds. I thought that some of the tests were being skipped but they weren't.

jomey:
> > Really nice improvement to the readability of the radiation component!
> > One thought I want to throw out there is to add a ticket to improve performance by vectorizing a lot of those functions? Without looking much into it, most calculations look like they can be easily converted with numba for instance.
> 
> I've tried to keep them as vectorized as possible to take advantage of numpy. Numba could definitely help on the larger basins or high resolution DEM's perhaps. As it stands, the straight python versions of these functions are 20x faster than the IPW versions. A 4 month simulations I did for a 500x1000 domain completed in like an hour instead of 3-4 hours (and not even threaded). So we're already seeing huge speed ups from this.
> 
> The tests freaked me out a little bit because the tests went from ~90 seconds to ~13 seconds. I thought that some of the tests were being skipped but they weren't.

That is a really good boost. Looking forward to apply that to ERW.
Part of the improvement probably is from removing all the calls to a subprocess, which can be expensive with many at once.

smrf/envphys/solar/cloud.py
```diff
@@ -0,0 +1,89 @@
...
+    basin_sol = df_solar.copy()
+    for idt, dt in enumerate(dates):
+        # get solar using twostream
+        dtt = pd.to_datetime(dt)
+        basin_sol.iloc[idt, :] = model_solar(dtt, lat, lon)
+
+    # if it's close to sun down or sun up, then the cloud factor gets
+    # difficult to calculate
+    basin_sol[basin_sol < 50] = 0
+    df_solar[basin_sol < 50] = 0
```
jomey, line 77: General question: Is 50 degrees consider sunset?

scotthavens:
And many of the IPW functions have a lot of file IO. For stoporad there are probably 6 or so temporary files that are generated and removing the IO can be fairly significant.

smrf/envphys/solar/cloud.py
scotthavens: This is 50 W/m2. There can be some unintended artifacts at low solar values for calculating cloud factor (measured / model) especially if the measured doesn't quite go to 0 at the same time as the modeled.

PR #176, opened 2020-07-02
Threading cleanup and testing

scotthavens:
Trying to tackle the larger issue #114. Added a specific Lakes test that uses gridded data and threading. There were a few things that needed to be fixed and I have moved all the thread variables to their modules for better control. Also added tests for outputting specific variables from both the threaded and non threaded case.


scotthavens:
There are some odd problems with the list copy for the thread variables. Might have to use the bigger `deepcopy` guns...

smrf/distribute/solar.py
```diff
@@ -243,6 +248,10 @@ def initialize(self, topo, data):
         self.veg_tau = topo.veg_tau
         self.veg_k = topo.veg_k
 
+        # self.thread_variables = self.CLEAR_SKY_THREAD_VARIABLES + \
+        #     self.CLOUD_THREAD_VARIABLES + self.VEG_THREAD_VARIABLES + \
+        #     ['net_solar']
+
```
jomey, line 254: Leftover from testing?

smrf/distribute/thermal.py
```diff
@@ -229,8 +234,8 @@ def initialize(self, topo, data):
         """
 
         self._logger.debug('Initializing distribute.thermal')
-        if self.gridded:
-            self._initialize(topo, data.metadata)
+        # if self.gridded:
```
jomey, line 237: Same here?

smrf/distribute/soil_temp.py
```diff
@@ -37,6 +39,8 @@ class ts(image_data.image_data):
     # be written during main distribute loop
     post_process_variables = {}
 
+    BASE_THREAD_VARIABLES = []
```
jomey, line 42: No need to declare the empty variable again, since it is declared in the parent class `image_data`.
There are more cases below.

smrf/framework/model_framework.py
```diff
@@ -217,6 +202,11 @@ def _setup_date_and_time(self):
         ))
         self.time_steps = len(self.date_time)
 
+    @property
+    def thermal_netcdf(self):
+        return self.distribute['thermal'].gridded and \
+            self.config['gridded']['data_type'] in ['wrf', 'netcdf']
```
jomey, line 208: `['wrf', 'netcdf']` smells like a good candidate for a class constant.

smrf/distribute/precipitation.py
```diff
@@ -164,6 +173,7 @@ def initialize(self, topo, data):
             """ model:  """.format(self.nasde_model))
 
         if self.nasde_model == 'marks2017':
+            self.thread_variables += ['storm_id']
```
jomey, line 176: Mainly a style preference and I tend to use `.append()` for adding to an array. Works for numpy arrays, for instance, as well.

smrf/distribute/solar.py
scotthavens: This was the original location for it before I moved it to a class constant.

smrf/distribute/thermal.py
scotthavens: We were only initializing if it was gridded but now it's done each time to get the base thread variables. Cleaned up

smrf/distribute/soil_temp.py
scotthavens: Good catch. Soil temp isn't added to the queue like others since it's constant.

smrf/framework/model_framework.py
scotthavens: This method was for the original thermal implementation where we're using the thermal from the atmospheric model. We've switched that with HRRR to calculated clear sky thermal and I'd like to see all the other gridded datasets do that as well to keep it consistent.

smrf/distribute/soil_temp.py
scotthavens: Actually this is needed as `_initialize` isn't called for soil temp.

smrf/distribute/precipitation.py
scotthavens: Used append everywhere else but here...

smrf/distribute/soil_temp.py
jomey: Any thoughts on this? I forgot to hit the final 'comment' button. Hopefully you received a notification by now.

scotthavens: These are tough because while they are constant, some variables need to be added to the queue dependent on configuration. So the problem I was running into in the tests if you add a variable to the base variable list, it will modify the class and not the instance. That's why I call `_intialize` to create another property that can be modified on the instance. Open to other ideas though as it still seems to be slightly problematic.

jomey: Can we merge this onto a branch instead, so I can do a quick sample implementation? Otherwise I would have to fork your fork to do this.

scotthavens: It's up on the main repo.

scotthavens:
I think that there are some thread variables that are persisting between test runs, that was the problem I was having especially going from station runs to HRRR runs.

jomey:
> I think that there are some thread variables that are persisting between test runs, that was the problem I was having especially going from station runs to HRRR runs.

Curious on why that only happens on the Python 3.6 runs. Last night, I was trying to see if there were some changes in Python and could not find any. I will have another look. Locally they consistently pass with 3.8

scotthavens:
It's not just the 3.6, I've seen it pass there then fail on OSX 3.8. So there is something that is persisting between the simulations in the queue

jomey:
I did some more reworking of the thread variable handling https://github.com/USDA-ARS-NWRC/smrf/tree/18_threading
to prevent convoluting the code with `deepcopy`

jomey:
I think we should close one of the branches. My vote would be for your fork, so we can both update.

scotthavens:
Sounds good, this one is puzzling!

scotthavens:
Closing in favor of #178

jomey:
I just updated the comparison for the time variable and it now passes. I am a little suspicious of that success though.
Can you run it on your local for a double check?

PR #177, opened 2020-07-08
ImageData - First step towards central thread variable.

jomey:
This is my first hunch. Looking further at the threading logic and consolidate the structure for non- vs distributed execution could simplify a few more pieces. Work in progress.

### DIFF
--- a/smrf/distribute/image_data.py
+++ b/smrf/distribute/image_data.py
@@ -44,6 +44,8 @@ class image_data():
 
     """
 
+    BASE_THREAD_VARIABLES = []
+
     def __init__(self, variable):
 
         self.variable = variable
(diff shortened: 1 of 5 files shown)


scotthavens:
The `@property` is probably cleaner way to do this. The thread variables are quite a mess at the moment but a lot of that won't be cleaned up until we clean up some of the distribute function calls. For example, `precip` requires a lot of thread variables even though not all are used. Although cleaning up `precip` is a whole other beast... which will probably happen once Ernesto precip rescaling is ready to be incorporated.

jomey:
The failing tests are the same that are failing on the original `18_threading` branch. Should we just merge this and take care of those on the original?

scotthavens:
So merge your branch into `18_threading`?

jomey:
> So merge your branch into `18_threading`?

Yup, if you are OK with this approach. The target for this PR is set to that. Only need to allow merging with not passed checks somehow

PR #178, opened 2020-07-09
Threading and clean up take 2

scotthavens:
Trying to tackle the larger issue #114. Added a specific Lakes test that uses gridded data and threading. There were a few things that needed to be fixed and I have moved all the thread variables to their modules for better control. Also added tests for outputting specific variables from both the threaded and non threaded case.


jomey:
I am a little suspicious that the test passed after using the numpy `assertEqual`. I wanted to get more details instead of a 'random' pass

scotthavens:
That is a little odd but probably a better comparison to use numpy which will compare the time values, not just the length. The way you set the hidden variable is good way to do this, especially with the frozenset.

The build will do a pretty good amount of testing, we can also retrigger the build another few times just to make sure it isn't a fluke.

PR #179, opened 2020-07-09
Input data overhaul

scotthavens:
A big issue with SMRF and gridded data is that it must loaded all upfront which does not make sense for a year long run. Refactored the data loading to be more organized and have an easier call in `SMRF`.

- The HRRR data can be loaded either upfront or each timestep through a new config option
- When running threaded, the data is loaded into a data queue that the threads get from
- All gridded data is used in the same way for thermal radiation, meaning that the clear sky thermal is modeled then adjusted for cloud factor and not distributing thermal radiation from the gridded dataset.
- TODO remove MySQL


jomey:
Jumping the gun here and I would suggest to get rid of the 'Load` and replace with 'Input` prefix for the the data classes.
Think it reads better, when doing something like this: `InputCSV().load` over `LoadCSV().load`

jomey:
I am also a fan of using the `__all__` in `__init__.py` files to make nicer import statements:
https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

scotthavens:
> Jumping the gun here and I would suggest to get rid of the 'Load` and replace with 'Input` prefix for the the data classes.
> Think it reads better, when doing something like this: `InputCSV().load` over `LoadCSV().load`

Can't you see this is a draft!?  😉

I'm torn whether or not to just start putting things in the imports one at a time or do it all at once in a separate PR dedicated to it. Then the class names could be fixed along with that.

jomey:
> 
> Can't you see this is a draft!? 😉
> 
It's like putting a Christmas present early under the tree and showing discipline not to peak into it :-)

> I'm torn whether or not to just start putting things in the imports one at a time or do it all at once in a separate PR dedicated to it. Then the class names could be fixed along with that.

Dedicated PR works to go through the import structure.

scotthavens:
> I am also a fan of using the `__all__` in `__init__.py` files to make nicer import statements:
> https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

It looks like this really only applies for the import using `from module import *`? I know that there is some issues with importing packages in SMRF. For example, `from smrf import utils` doesn't load the packages in `utils` but `import smrf.utils` does.

jomey:
> It looks like this really only applies for the import using `from module import *`? I know that there is some issues with importing packages in SMRF. For example, `from smrf import utils` doesn't load the packages in `utils` but `import smrf.utils` does.

Yes, I double checked that the other day as well and was a bit dissapointed since I always understood that this controls which componentes you expose to the import API.

For your examples, I think the first example should work this way:
```python
from smrf import utils
utils.method
```
No?

scotthavens:
That's how I would expect it to work as well. I'm not sure if it was in a development environment where it wasn't installed properly but I've scratched my head a few times when that happens.

scotthavens:
This PR is no longer updating with new pushes, closing for a new one.

PR #181, opened 2020-07-10
Rework test suite base class

jomey:
Sneaked quite some test structure changes into here and this should set a good foundation for the suite going forward.

Significant change is two fold:
1. Rework the `SMRFTestCase` base class to use more methods and properties to enable easier overwrite for different config setups. This also added a method to copy the base config to ensure the initial copy will not be changed.
2. Moved the two sample data basins into a `basins` folder. This separates test data from actual test logic.


smrf/tests/test_configurations.py
```diff
@@ -1,13 +0,0 @@
-from smrf.framework.model_framework import can_i_run_smrf
-from smrf.tests.smrf_test_case import SMRFTestCase
-
-
-class TestConfigurations(SMRFTestCase):
```
jomey, line 5: Made the decision to remove this spec since I did not see much value. Other tests will start to fail and this one would not stand out.

smrf/tests/smrf_test_case.py
```diff
@@ -15,17 +17,70 @@ class SMRFTestCase(unittest.TestCase):
...
 
-    def can_i_run_smrf(self, config):
+    @classmethod
+    def remove_output_dir(cls):
+        if hasattr(cls, '_output_dir') and \
+                os.path.exists(cls._output_dir):
+            shutil.rmtree(cls._output_dir)
+
+    @staticmethod
+    def can_i_run_smrf(config):
```
jomey, line 83: At some point I would like to see this no longer in use.

scotthavens: Agreed and also remove it from `framework`

smrf/tests/output/test_output_variables.py
```diff
@@ -60,9 +60,9 @@ def setUpClass(cls):
         """
         super().setUpClass()
 
-        config = copy(cls.base_config)
+        config = cls.base_config_copy()
```
scotthavens, line 63: Handy!

smrf/tests/smrf_test_case.py
```diff
@@ -60,6 +117,9 @@ def assertGoldEqual(self, gold, not_gold, error_msg):
                 err_msg=error_msg
             )
 
+    def setUp(self):
+        self._dist_variables = None
```
scotthavens, line 121: What is the thinking for setting this to `None` in order to copy the `DIST_VARIABLES`?

smrf/tests/smrf_test_case.py
```diff
@@ -15,17 +17,71 @@ class SMRFTestCase(unittest.TestCase):
...
         'precip',
         'thermal',
         'vapor_pressure',
         'wind',
-    ]
+    ])
+
+    BASE_INI_FILE_NAME = 'config.ini'
+
+    test_dir = PurePath(smrf.__file__).parent.joinpath('tests')
```
scotthavens, line 34: Per your preference, why not capitalize these?

smrf/tests/smrf_test_case_lakes.py
```diff
@@ -1,47 +1,18 @@
...
-import smrf
 from smrf.tests.smrf_test_case import SMRFTestCase
 
 
 class SMRFTestCaseLakes(SMRFTestCase):
     """
-    The base test case for SMRF that will load in the configuration file
-    and store as the base config. Also will remove the output directory
-    upon tear down.
+    Runs the short simulation over reynolds mountain east
```
scotthavens, line 8: Lakes?

smrf/tests/test_configurations.py
scotthavens: Never a fan of this test, it was always pretty pointless.

smrf/tests/test_data_gridded.py
```diff
@@ -10,21 +9,19 @@
 
 class TestLoadGrid(SMRFTestCase):
 
-    def compare_hrrr_gold(self, out_dir):
+    def compare_hrrr_gold(self):
```
scotthavens, line 12: Could this be put on the base class? We have 2 methods for comparing gold files, this one and the `compare_netcdf_files` in the `SMRFTestCase`

smrf/tests/test_lakes_wind_ninja.py
```diff
@@ -12,98 +11,86 @@
 
 class TestLakes(SMRFTestCaseLakes):
     """
-    Integration test for SMRF using reynolds mountain east without threading
+    Integration test for SMRF without threading.
+    Runs the short simulation over reynolds mountain east.
```
scotthavens, line 15: Lakes?

smrf/tests/test_lakes_wind_ninja.py
```diff
@@ -12,98 +11,86 @@
...
-            os.path.join(self.output, 'vapor_pressure.nc')
+            os.path.join(self.output_dir, 'vapor_pressure.nc')
         )
 
 
 class TestLakesThreaded(SMRFTestCaseLakes):
-
+    """
+    Integration test for SMRF.
+    Runs the short simulation over reynolds mountain east.
```
scotthavens, line 87: Lakes?

smrf/tests/smrf_test_case.py
jomey: https://github.com/USDA-ARS-NWRC/smrf/issues/182

smrf/tests/smrf_test_case.py
jomey: I kept this lower case to give it the same accessing syntax as you would do it by declaring a `@property` for an instance. All caps symbolizes a constant (to me). This can be argued as personal preference and I can also change it to use all caps.

What do you prefer?

smrf/tests/smrf_test_case.py
jomey: By ensuring that all instances have a internal variable declared, you can do this kind of 'lazy loading' with a property:
```
    @property
     def dist_variables(self):
         if self._dist_variables is None:
```
Otherwise you could go the route of `getattr()`.

Preference?

smrf/tests/smrf_test_case_lakes.py
jomey: Brainless copying (see former line 19)....

smrf/tests/smrf_test_case.py
scotthavens: Works for me

smrf/tests/smrf_test_case.py
scotthavens: This works fine, it's not really a constant since it's dynamically set so makes sense to have it lower case.

smrf/tests/test_lakes_wind_ninja.py
```diff
@@ -12,101 +9,59 @@
...
-    def tearDown(self):
-        pass
-
     def test_air_temp(self):
-
-        self.compare_netcdf_files(
-            os.path.join(self.gold, 'air_temp.nc'),
-            os.path.join(self.output, 'air_temp.nc')
-        )
+        self.compare_netcdf_files('air_temp.nc')
```
scotthavens, line 25: Clean!

jomey:
Did some updates after resolving all the merge conflicts with latest master. Mostly organization.

smrf/tests/smrf_test_case.py
```diff
@@ -15,17 +17,87 @@ class SMRFTestCase(unittest.TestCase):
...
         'vapor_pressure',
         'wind',
-    ]
+    ])
+
+    THREAD_CONFIG = {
+        'threading': True,
+        'max_queue': 1,
+        'time_out': 5
+    }
```
jomey, line 36: Central config for test with threading

smrf/tests/test_threading.py
```diff
@@ -1,15 +1,13 @@
...
 
 
 class TestThreading(SMRFTestCaseLakes):
+    THREAD_VARIABLE_COUNT = 36
 
-    def get_variables(self, config):
-        s = SMRF(self.base_config)
+    @staticmethod
+    def get_variables(config):
+        s = SMRF(config)
```
jomey, line 10: Think there was a bug here, where the passed in config was not used and the base without threading instead.

scotthavens: I was attempting to run it not threaded first then threaded to see if any of the variables changed.

jomey: Should we keep the test or nuke this class?

scotthavens: I don't know, I'm on the fence. It's useful in the sense it makes us aware of what is in the queue. But it's a lot of extra code for that check.

Or can we move the `assert_thread_variables` here for a quick test?

jomey: That's sound like a perfect place for it.

smrf/tests/test_threading.py
```diff
@@ -1,36 +1,75 @@
...
+from smrf.framework.model_framework import SMRF
 from smrf.tests.smrf_test_case_lakes import SMRFTestCaseLakes
 
 
 class TestThreading(SMRFTestCaseLakes):
-    THREAD_VARIABLE_COUNT = 36
+    """
+    Test for all defined BASE_THREAD_VARIABLES, not accounting for config
+    dependent additions.
+    """
```
jomey, line 9: This is what I came up with

scotthavens: Looks good

Issue #180, opened 2020-07-10
Intermittent test failures for wind


jomey:
Tests wind in a threaded setup fail intermittently with a missing time step.
Example: https://travis-ci.com/github/USDA-ARS-NWRC/smrf/jobs/359503602

scotthavens:
This probably still has something to do with PR #178 where the thread variables for the wind class aren't being set properly so some of them persist from test to test.

jomey:
I am wondering, why the time comparison has one less hour than expected. How are these related with the thread variables?

scotthavens:
Not all the thread variables for the Winstral model are used for Wind Ninja. However, they still need to be set for precipitation so if they aren't set then they can fail. It typically fails at 2 timesteps because that is the max number in the queue. Changing max queue to 1 would make it fail at 1 I bet.

scotthavens:
Fixed with #183

PR #183, opened 2020-07-13
Wind thread variables clean up

scotthavens:
Trying to clean up the thread variables for the wind variables to address #180


smrf/distribute/image_data.py
```diff
@@ -62,6 +62,18 @@ def thread_variables(self):
...
+        """Add a list or single variable to the thread variables
+
+        Args:
+            variables (list or str): List or string of variables to add
+        """
+
+        if isinstance(variables, str):
+            variables = list(variables)
+
+        self._thread_variables = self.thread_variables + variables
```
jomey, line 75: I think this will cause an issue, where the method will be a one time use and successive of the method will 'forget' previously added ones.

I think going with a property setter and a separate add method is the way to go:
```python
    @thread_variables.setter
    def thread_variables(self, value):
        self._thread_variables = value

    def add_thread_variable(self, variables):
        if isinstance(variables, str):
            variables = list(variables)
            
        self.thread_variables = self.thread_variables + variables
```

Something like this (untested code)...

smrf/distribute/wind/wind.py
```diff
@@ -111,10 +120,14 @@ def initialize(self, topo, data):
...
 
         self._logger.debug('Initializing distribute.wind')
-
+        self.date_time = date_time
         self.wind_model._initialize(topo, data.metadata)
 
-        if self.config['wind_model'] != 'interp':
+        if self.check_wind_model_type('winstral'):
+            self.add_thread_variables(
+                list(self.wind_model.BASE_THREAD_VARIABLES))
```
jomey, line 128: The first access  of `self.wind_model.thread_variables` should do this for you without the need to set this beforehand.

smrf/distribute/image_data.py
scotthavens: Makes sense, the setter works

smrf/distribute/wind/wind.py
scotthavens: Moved this to a property of the wind model

scotthavens:
It's still happening...

smrf/distribute/wind/winstral.py
```diff
@@ -107,6 +113,10 @@ def __init__(self, smrf_config):
 
         self.veg = v
 
+    @property
+    def thread_variables(self):
+        return list(self.BASE_THREAD_VARIABLES)
+
```
jomey, line 119: Is this required? What breaks, when not overwriting and using the definition from the `image_data` class?

smrf/distribute/wind/wind.py
```diff
@@ -81,22 +78,34 @@ def __init__(self, config):
...
-        if self.config['wind_model'] == 'interp':
+        if self.check_wind_model_type('interp'):
             # Straight interpolation of the wind
             self.wind_model = self
             self.wind_model.flatwind = None
             self.wind_model.cellmaxus = None
             self.wind_model.dir_round_cell = None
 
-        elif self.config['wind_model'] == 'wind_ninja':
+        elif self.check_wind_model_type('wind_ninja'):
```
jomey, line 88: For future improvements, I would declare a constants in each class (`WindNinja` and `Windstral`) and then pass this as an argument. There are several other comparisons in this  class against this string and it would be nice to have one central definition of that.

smrf/distribute/wind/winstral.py
scotthavens: Because I had in my mind that this didn't extend the `image_data` class, but it does...

jomey:
> It's still happening...

Was that locally? The error was from Coverall for the checks.

smrf/framework/model_framework.py
```diff
@@ -620,18 +629,20 @@ def create_distributed_threads(self):
         self._logger.info("Initializing distributed variables...")
 
         # These are the variables that will be queued
-        thread_queue_variables = list(self.BASE_THREAD_VARIABLES)
```
jomey, line 623: One of my thoughts here was to keep the `thread_queue_variables` to the local method scope and not have it set on the instance. That was a guess though on where it potentially could persist across runs.

scotthavens:
> > It's still happening...
> 
> Was that locally? The error was from Coverall for the checks.

It happened on the first run for 3.6, I restarted the job and then it failed on coveralls. [This one has failed](https://travis-ci.com/github/scotthavens/smrf/builds/175437260)

smrf/framework/model_framework.py
scotthavens: It seems that this probably isn't where things are being persisted across runs if it's still happening. I've set it on the class to try and see if it's changing in length. Might have to be a little more aggressive in the tests to see where it's failing.

scotthavens:
The latest run failed again but it did not fail on the thread variables. So it appears that the thread variables are being set properly. This may point to the failure coming from the output or cleaner thread.

scotthavens:
It's seem to have been fixed. I have ran the latest tests 3 times without failure (except for coveralls which is allowed to fail now). While the output queues were cleaned up, I think the actual fix was increasing the timeout to 5 seconds... I don't thing the Travis VM's are all that fast so the queues were timing out while SMRF was running.

smrf/distribute/image_data.py
```diff
@@ -62,6 +63,30 @@ def thread_variables(self):
...
+
+    def add_thread_variables(self, variables):
+        """Add a list or single variable to the thread variables
+
+        Args:
+            variables (list or str): List or string of variables to add
+        """
+
+        if isinstance(variables, str):
+            variables = list((variables, ))
```
jomey, line 78: When I first saw this, I was curious of there is an alternative to this approach.
SO suggested: `variable.split()`

Just for reference

scotthavens: So if it was a list already, then it would just output another list?

smrf/tests/Lakes/config.ini
```diff
@@ -124,4 +124,4 @@ input_backup:                  True
 [system]
 threading:                     False
 log_level:                     debug
-log_file:                      ./output/log.txt
+#log_file:                      ./output/log.txt
```
jomey, line 127: No more logs in test runs?

smrf/tests/smrf_test_case_lakes.py
```diff
@@ -45,3 +45,71 @@ def tearDownClass(cls):
         folder = os.path.join(cls.base_config.cfg['output']['out_location'])
         if os.path.exists(folder):
             shutil.rmtree(folder)
+
+    def assert_list(self, list1, list2):
+        """Couldn't get the assert_list to work"""
+
+        self.assertTrue(sorted(list1) == sorted(list2))
+
+    def assert_thread_variables(self):
```
jomey, line 54: Would this be worth splitting into a test case for each distribution?

smrf/tests/Lakes/config.ini
scotthavens: Just for Lakes, so 2 runs. It's to try and catch any useful output if it does fail in threading. If the issue doesn't crop back up then we can change it back.

smrf/tests/smrf_test_case_lakes.py
scotthavens: The RME case will have two separate cases of thread variables, one for stations and gridded. Since the Lakes only uses HRRR, it's a little easier problem. I don't know if its worth it to fully flush this out.

smrf/distribute/image_data.py
jomey: No, it's to cast a string to a list.
```python
'foo'.split()
['foo']

```

scotthavens: Ah, that would make it a bit cleaner.

smrf/tests/smrf_test_case_lakes.py
```diff
@@ -45,3 +45,71 @@ def tearDownClass(cls):
...
+
+        self.assert_list(
+            self.smrf.distribute['air_temp'].thread_variables,
+            ['air_temp']
```
jomey, line 58: Instead of defining a new list here, you could use the defined values in the classes:
```
ta.BASE_THREAD_VARIABLES
```

This could ensure that we are testing, what we are defining and enqueuing.

smrf/tests/test_lakes_wind_ninja.py
```diff
@@ -21,7 +21,10 @@ def setUpClass(cls):
         """
         super().setUpClass()
 
-        run_smrf(cls.config_file)
+        cls.smrf = run_smrf(cls.config_file)
+
+    def setUp(self):
+        self.assert_thread_variables()
```
jomey, line 27: Is this to ensure an empty start?
Maybe not check for every test case?

scotthavens: I was having trouble with it in the `setupClass` so I moved it here. It's not going to change from test to test.

smrf/tests/smrf_test_case_lakes.py
scotthavens: Makes sense for some of the easier ones. But this test will fail when a new variable is added which may ensure that we go through this process again to ensure that it's being properly added to the variable list.

jomey:
> It's seem to have been fixed. I have ran the latest tests 3 times without failure (except for coveralls which is allowed to fail now). While the output queues were cleaned up, I think the actual fix was increasing the timeout to 5 seconds... I don't thing the Travis VM's are all that fast so the queues were timing out while SMRF was running.

The timeout sounds very plausible. CI environments are prone to fluctuating test execution times depending on their load.
On one other project I work on, the test times for the (very big) suite fluctuates between 70 and 180 minutes.

scotthavens:
> > It's seem to have been fixed. I have ran the latest tests 3 times without failure (except for coveralls which is allowed to fail now). While the output queues were cleaned up, I think the actual fix was increasing the timeout to 5 seconds... I don't thing the Travis VM's are all that fast so the queues were timing out while SMRF was running.
> 
> The timeout sounds very plausible. CI environments are prone to fluctuating test execution times depending on their load.
> On one other project I work on, the test times for the (very big) suite fluctuates between 70 and 180 minutes.

Made some good improvements even it the execution time was the problem!

Issue #127, opened 2019-11-25
MySQL data deprecation?


scotthavens:
Should SMRF deprecate the MySQL code moving forward? This isn't something that is used much anymore due to the switch to gridded data and that none of the MySQL projects (like the database and `CorrectWxData`) are being actively maintained.

Hedrick-ARS:
The way we set up the MySQL database was pretty ad hoc. We've always been skeptical that anyone using SMRF/AWSM would be able to implement a similar database, so I see no reason why we shouldn't deprecate the functionality since it was tailored to our use case. It seems much more likely that if someone wants to use stations for forcing input that they could just construct their own CSV files.

PR #184, opened 2020-07-15
Input data overhaul

scotthavens:
A big issue with SMRF and gridded data is that it must loaded all upfront which does not make sense for a year long run. Refactored the data loading to be more organized and have an easier call in SMRF.

- The HRRR data can be loaded either upfront or each timestep through a new config option
- When running threaded, the data is loaded into a data queue that the threads get from
- All gridded data is used in the same way for thermal radiation, meaning that the clear sky thermal is modeled then adjusted for cloud factor and not distributing thermal radiation from the gridded dataset.
- Remove MySQL for #127 
- Removed the `last_storm_basin` metric what calculated the time since last storm for the whole basin, this is not being used anymore


docs/getting_started/run_smrf.rst
```diff
@@ -30,7 +30,7 @@ below is the function :mod:`run_smrf <smrf.framework.model_framework.run_smrf>`.
         s.loadTopo()
 
         # initialize the distribution
-        s.initializeDistribution()
+        s.create_distribution()
```
scotthavens, line 33: This will break AWSM. Should be included in [AWSM #68](https://github.com/USDA-ARS-NWRC/awsm/issues/68)

smrf/data/load_topo.py
```diff
@@ -104,18 +104,11 @@ def readImages(self, f):
         # to double or int
         for v_smrf in self.IMAGES:
 
-            # check to see if the user defined any variables
-            # e.g. veg_height = veg_length
```
scotthavens, line 108: This has never been used where one can specify different field names in the topo.nc. Most will be generated with `basin_setup` so it's just making it a little more rigid.

smrf/data/wrf.py
```diff
@@ -0,0 +1,188 @@
...
+        self.wind_speed = pd.DataFrame(index=time, columns=self.primary_id)
+        self.wind_direction = pd.DataFrame(index=time, columns=self.primary_id)
+        min_speed = 0.47
+
+        u10 = self.file.variables['UGRD'][time_ind, :]
+        v10 = self.file.variables['VGRD'][time_ind, :]
+
+        # calculate the wind speed
+        s = np.sqrt(u10**2 + v10**2)
+        s[s < min_speed] = min_speed
```
jomey, line 174: Same as with loading HRRR data. Feels like this can be done with one call instead of two

smrf/distribute/air_temp.py
```diff
@@ -67,6 +67,7 @@ def initialize(self, topo, data):
         """
 
         self._logger.debug('Initializing distribute.air_temp')
+        self.date_time = date_time
```
jomey, line 70: If something is an optional parameter to a method, but with a default value, I tend to use this:
```python
def initialize(self, topo, data, **kwargs):
            self.date_time = kwargs.pop('date_time', None)
```

Opens up the ability to add more parameters without having to touch the method signature and calling methods can be found easier, that use the optional parameter.

smrf/distribute/solar.py
```diff
@@ -243,6 +243,7 @@ def initialize(self, topo, data):
         """
 
         self._logger.debug('Initializing distribute.solar')
+        self.date_time = date_time
         # Solar has no stations. Relies on Cloud factor
         self.stations = None
         self._initialize(topo, data.metadata)
```
jomey, line 249: Looks like 'initialize' should be in the `image_data.image_data` and calls `self._initialize` then all inheriting classes call `super()` and add their customization

smrf/tests/data/test_wrf.py
```diff
@@ -0,0 +1,59 @@
...
+        config = cast_all_variables(config, config.mcfg)
+
+        # ensure that the recipes are used
+        self.assertTrue(
+            'station_adjust_for_undercatch' not in config.cfg['precip'].keys()
+        )
+        self.assertFalse(config.cfg['thermal']['correct_cloud'])
+        self.assertTrue(config.cfg['thermal']['correct_veg'])
+
+        run_smrf(config)
```
jomey, line 59: Why is SMRF run again, but nothing verified after?

smrf/tests/data/test_netcdf.py
```diff
@@ -0,0 +1,55 @@
...
+        config.apply_recipes()
+        config = cast_all_variables(config, config.mcfg)
+
+        # ensure that the recipes are used
+        self.assertTrue(
+            'station_adjust_for_undercatch' not in config.cfg['precip'].keys())
+        self.assertFalse(config.cfg['thermal']['correct_cloud'])
+        self.assertTrue(config.cfg['thermal']['correct_veg'])
+
+        run_smrf(config)
```
jomey, line 55: Same question - SMRF is run and nothing verified

smrf/utils/utils.py
```diff
@@ -448,3 +447,36 @@ def grid_interpolate_deconstructed(tri, values, grid_points, method='linear'):
...
+    an increment
+    """
+    result = []
+    nxt = start_date
+
+    while nxt <= end_date:
+        result.append(nxt)
+        nxt += increment
+
+    return np.array(result)
```
jomey, line 482: I feel that I just recently removed this method in favor of 
```python
list(pd.date_range(
            self.start_date,
            self.end_date,
            freq="{[time][time_step]}min".format(self.config),
            tz=self.time_zone
        ))
```

jomey:
Some of my comments apply to multiple data loading classes and I did not repeat the message there. Easier to discuss suggestions in one place.

smrf/data/csv.py
```diff
@@ -0,0 +1,76 @@
...
+
+from smrf.utils.utils import check_station_colocation
+
+
+class InputCSV():
+
+    def __init__(self, *args, **kwargs):
+
+        for keys in kwargs.keys():
+            setattr(self, keys, kwargs[keys])
```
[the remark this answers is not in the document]
scotthavens: Agreed. At some point there could be a base class for the input data.

smrf/data/csv.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self._logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
scotthavens: It was mainly for ease of use within a module, just create it in the `__init__` then use it everywhere. Then it's also logging what module the log line is coming from.

smrf/data/hrrr_grib.py
```diff
@@ -0,0 +1,179 @@
...
+        self._logger = logging.getLogger(__name__)
+
+        if self.config['hrrr_load_method'] == 'timestep':
+            self.timestep_dates()
+            self.cf_memory = None
+
+        self.hrrr = HRRR(external_logger=self._logger)
+
+    def timestep_dates(self):
+        self.end_date = self.start_date + pd.to_timedelta(20, 'minutes')
```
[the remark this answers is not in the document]
scotthavens: Good call

smrf/data/hrrr_grib.py
```diff
@@ -0,0 +1,179 @@
...
+        self.metadata = metadata
+
+        idx = data['air_temp'].index
+        cols = data['air_temp'].columns
+
+        self._logger.debug('Loading air_temp')
+        self.air_temp = data['air_temp']
+
+        # calculate vapor pressure
+        self._logger.debug('Loading vapor_pressure')
```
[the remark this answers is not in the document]
scotthavens: To keep you on your toes!

smrf/data/hrrr_grib.py
```diff
@@ -0,0 +1,179 @@
...
+            data['air_temp'].values,
+            data['relative_humidity'].values)
+        self.vapor_pressure = pd.DataFrame(vp, index=idx, columns=cols)
+
+        # calculate the wind speed and wind direction
+        self._logger.debug('Loading wind_speed and wind_direction')
+        min_speed = 0.47
+
+        s = np.sqrt(data['wind_u']**2 + data['wind_v']**2)
+        s[s < min_speed] = min_speed
```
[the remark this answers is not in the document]
scotthavens: It's actually not needed as the wind module will set the minimum wind speed

smrf/data/load_data.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # load the data
+        self.load_class = self.DATA_FUNCTIONS[self.data_type](**data_inputs)
+        self.load_class.load()
+
+        self.set_variables()
+
+        self.metadata_pixel_location()
+
+        if self.data_type == 'csv':
```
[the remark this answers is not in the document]
scotthavens: Makes sense, but wouldn't it be `InputCSV.DATA_TYPE`?

smrf/data/csv.py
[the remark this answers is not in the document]
jomey: I was referring to the naming with the leading "_" vs "self.logger"

smrf/data/load_data.py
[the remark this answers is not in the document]
jomey: Yup, I can see this being treated as a constant over an attribute.

smrf/data/load_data.py
```diff
@@ -0,0 +1,183 @@
...
+            d = getattr(self.load_class, variable, None)
+            if variable == 'metadata':
+                setattr(self, variable, d)
+            elif d is not None:
+                d = d.tz_convert(tz=self.time_zone)
+                setattr(self, variable, d[self.start_date:self.end_date])
+
+    def metadata_pixel_location(self):
+        """Set the pixel location in the topo for each station
+        """
```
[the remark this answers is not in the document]
scotthavens: Yes, apply is slow and this could be sped up at some point.

smrf/data/csv.py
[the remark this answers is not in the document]
scotthavens: Not sure the reason I put the `_` in there initially, probably came from an example off SO or medium...

smrf/data/wrf.py
scotthavens: Same as above, we'll let the wind module handle this

smrf/distribute/air_temp.py
scotthavens: Cool!

scotthavens: But along the same lines as the data, it's somewhat tough to know what is expected for the method.

smrf/distribute/solar.py
scotthavens: Makes sense, might be a bit of scope creep on this PR though

smrf/tests/data/test_wrf.py
scotthavens: It's more of an end to end test to make sure that we can load the WRF data and run SMRF. We never created gold files for this and the netcdf run because we don't use these as often.

smrf/utils/utils.py
scotthavens: Haha, you did. Moved your function to utils to allow it to be reused.

scotthavens:
> Some of my comments apply to multiple data loading classes and I did not repeat the message there. Easier to discuss suggestions in one place.

Got it, a lot of the classes are similar so I've applied the comments to all classes that were applicable.

smrf/distribute/air_temp.py
jomey: Sounds like another PR/ticket to look into details for this.

smrf/tests/data/test_wrf.py
jomey: So mainly make sure nothing blows up?

smrf/distribute/solar.py
jomey: #185

smrf/tests/data/test_wrf.py
scotthavens: Exactly and to test that the load functions actually load the data.

smrf/data/load_data.py
```diff
@@ -0,0 +1,184 @@
...
+        'vapor_pressure': 'vapor_pressure',
+        'precip': 'precipitation',
+        'wind_speed': 'wind',
+        'wind_direction': 'wind',
+        'cloud_factor': 'cloud_factor',
+        'thermal': 'thermal'
+    }
+
+    DATA_FUNCTIONS = {
+        'csv': InputCSV,
```
jomey, line 45: I guess you could use the data type now - ` InputCSV.DATA_TYPE`

scotthavens: Yep, updated these as well.

smrf/data/load_data.py
```diff
@@ -0,0 +1,184 @@
...
+        self.time_zone = start_date.tzinfo
+        self.topo = topo
+
+        self._logger = logging.getLogger(__name__)
+
+        # get the buffer gridded data domain extents in lat long
+        self.model_domain_grid()
+
+        if 'csv' in self.smrf_config:
+            self.data_type = 'csv'
```
jomey, line 68: Would it make sense to use the ` InputCSV.DATA_TYPE` here too?

scotthavens: My thoughts would be that the first `csv` refers to the config option, but the second could use it.

PR #187, opened 2020-09-17
169 numpy

scotthavens:
A 2 part PR.

## Numpy version

Numpy > v1.19 made some underlying changes that made the vapor pressure calculation slightly different only from HRRR. Looking at the vapor pressure calculation, the air temperature and relative humidity were exactly the same. The differences in vapor pressure stem from the `sati` and `satw` with differences in the float tolerance range.

Added a few levels of tests. First still tests for exact match. If the arrays are not exact, then try `allclose` with relative tolerance of 1e-5 and 1e-4. Raise an exception if these are not met.

## SMRF pip install

The setup.py imports numpy which will not work if numpy is not installed. Following [steps here](https://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py), moved the build_ext to a class so numpy is only imported if compiling the pyx files to C.

This does move away from using Cython build_ext but not fully sure that it is needed as setuptools will call gcc to compile.


scotthavens:
Added Cython `build_ext` back in. If Cython is present, it will compile SMRF from the .pyx files. If not, it willl use the .c files to build the extensions.

notebooks/numpy_differences.ipynb
```diff
@@ -0,0 +1,121 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": "array([-2.11303719e-04, -2.09472666e-04, -7.00683595e-05, -6.70166046e-05,\n       -5.78613290e-05, -1.80358875e-05,  0.00000000e+00,  1.29699711e-04,\n        1.82006839e-04,  1.91406260e-04])"
```
scotthavens, line 99: These are the differences in vapor pressure from the HRRR values.

setup.py
```diff
@@ -3,30 +3,40 @@
...
 
-with open('requirements.txt') as requirements_file:
-    requirements = requirements_file.read()
+class build_ext(_build_ext):
+    def finalize_options(self):
+        _build_ext.finalize_options(self)
+        # Prevent numpy from thinking it is still in its setup process:
+        __builtins__.__NUMPY_SETUP__ = False
+        import numpy
+        self.include_dirs.append(numpy.get_include())
```
Hedrick-ARS, line 31: This is hieroglyphics to me. I guess what I really would like to know is what is this improving, specifically?

scotthavens: Haha, me too. I just copied it from SO and it works. It's making setup.py load numpy if we're compiling the .pyx (cython) to C. This gets around an import numpy at the top, which will break smrf installing if numpy isn't already installed.

notebooks/numpy_differences.ipynb
jomey: So did you manually copy these to compare them?

requirements.txt
```diff
@@ -1,8 +1,7 @@
 coloredlogs
-Cython>=0.28.4
 inicheck>=0.9.0,<0.10.0
 netCDF4>=1.2.9
-numpy>=1.14.0,<1.19.0
+numpy>=1.14.0
```
jomey, line 4: Curious if a bigger than a specific version needs to be there? My understanding is that it will always take the latest and 1.14 was released in January of 2018. I would feel pretty confident at this point that it won't get anything older than that.

setup.py
Hedrick-ARS: Ok. I get it. So at what point does numpy get installed if it isn't already?

jomey: Where is the import? From within Cython?

smrf/tests/smrf_test_case.py
```diff
@@ -121,11 +121,25 @@ def assert_gold_equal(gold, not_gold, error_msg):
...
+                        not_gold,
+                        gold,
+                        atol=0,
+                        rtol=rtol
+                    )
+                    if status:
+                        break
+                if not status:
+                    raise AssertionError(error_msg)
+                print('Arrays are not exact match but close with rtol={}'.format(rtol))  # noqa
```
jomey, line 142: Is this status check needed?
Wouldn't it raise another `AssertionError` on the `allclose`?

jomey (approved):

requirements.txt
scotthavens: I did not manually copy, exported the data frame to csv then compared. So sorta manual.

Probably could do away with the minimum bound now.

setup.py
scotthavens: numpy will be installed as a dependency to SMRF, it's just not needed now to install SMRF, which was the hangup.

All numpy was needed for in setup.py was `numpy.get_include()` for it's headers. Other than that, it isn't needed. If we're compiling the .pyx to .c, then numpy will be used in the `build_ext`.

smrf/tests/smrf_test_case.py
scotthavens: I moved away from the `np.testing.allclose`. Now that I'm looking at it again, I can use the testing version.

scotthavens: Now I remember why! It may fail with the 1e-5 but not with the 1e-4. So I didn't want the test to fail until it made it all the way through.

Hedrick-ARS (approved):
Looks great to me.

Issue #163, opened 2020-06-10
Improve class names


jomey:
There are a few classes in SMRF, which do not follow PEP8 recommendations. For instance, `smrf/distribute/cloud_factor.py` uses `cl` as its class name.

This ticket is to go through the codebase and ensure that classes are easy to identify and follow PEP8 naming conventions.

PR #188, opened 2020-10-01
186 awsm integration

scotthavens:
There was a lot of duplicate code in AWSM from SMRF. This cleans up some of the variables or functions that need to be accessed from AWSM.

- `distribute_single_timestep()` method will distribute a single time step
- renamed some methods to be clearer about running in serial or threaded
- output variable dictionary that returns a dictionary
- can pass an external logger to `run_smrf`

### DIFF
--- a/smrf/framework/model_framework.py
+++ b/smrf/framework/model_framework.py
@@ -642,7 +642,7 @@ class SMRF():
 
             if output_variable in self.possible_output_variables.keys():
                 fname = join(out_location, output_variable)
-                module = self.possible_output_variables[output_variable]['module']
+                module = self.possible_output_variables[output_variable]['module']  # noqa
 
                 # TODO this is a hack to not have to redo the gold files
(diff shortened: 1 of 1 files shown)


smrf/framework/model_framework.py
```diff
@@ -668,8 +661,27 @@ def initializeOutput(self):
                 self._logger.error(
                     '{} not an output variable'.format(output_variable))
 
+        return variable_dict
+
+    def initializeOutput(self):
```
jomey, line 666: Rename to be consistent: `initialize_output`?

jomey (approved):

smrf/framework/model_framework.py
scotthavens: There are a few that are still camel case, like `loadTopo` and `loadData`. Perhaps might be better suited for #163 as some are used in many places.

PR #190, opened 2020-10-05
189 output 64-bit netcdf files

scotthavens:
Added an option to output SMRF netcdf files as float (32-bit) or double (64-bit). The default is float (32-bit) and the option for double will be used in AWSM when comparing all the different methods.

### DIFF
--- a/smrf/tests/output/test_netcdf.py
+++ b/smrf/tests/output/test_netcdf.py
@@ -1,6 +1,7 @@
 import os
 
 import netCDF4 as nc
+import numpy as np
 
 from smrf import __version__
 from smrf.framework.model_framework import SMRF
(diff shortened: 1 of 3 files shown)


smrf/tests/output/test_netcdf.py
```diff
@@ -10,14 +11,11 @@
 
 class TestOutputNetCDF(SMRFTestCase):
 
-    @classmethod
-    def setUpClass(cls):
+    def setUp(self):
         super().setUpClass()
-        cls.smrf = SMRF(cls.config_file)
+        self.smrf = SMRF(self.config_file)
```
jomey, line 16: Why did you need to switch to an instance here?

scotthavens: It gets initialized with the 64-bit test first and doesn't get overwritten. So a new instance was needed to set all the config

PR #193, opened 2020-10-15
Extra requires

jomey:


setup.py
```diff
@@ -139,13 +139,14 @@ def finalize_options(self):
     ],
     extras_require={
         'docs': [
-            'sphinxcontrib-bibtex',
-            'sphinxcontrib-websupport',
-            'pydata-sphinx-theme'
+            'Sphinx>=3.0,<=4.0',
+            'pydata-sphinx-theme',
```
jomey, line 143: Will this play nicely with readthedocs? Have no experience with that

README.md
```diff
@@ -37,7 +37,7 @@ git clone https://github.com/USDA-ARS-NWRC/smrf.git
 And install the requirements, SMRF and run the tests.
 
 ```bash
-python3 -m pip install -r requirements_dev.txt
+python3 -m pip install -r requirements_dev.txt .[tests]
```
jomey, line 40: Forgot to update that with the last PR

setup.py
scotthavens: Should, all versions of Sphinx are supported

scotthavens:
Can you add installing the extras into the [docker image](https://github.com/USDA-ARS-NWRC/smrf/blob/dbbaa87583502ceeda23aa1c6f618dfde0958348/Dockerfile#L37)? This also uses nose to test the container was built properly.

Dockerfile
```diff
@@ -34,9 +34,9 @@ COPY --from=builder /root/.local /usr/local
 
 RUN apt-get update -y \
     && apt-get install -y --no-install-recommends libeccodes-tools \
-    && python3 -m pip install --no-cache-dir nose \
+    && python3 -m pip install --no-cache-dir nose smrf[tests] \
     && nosetests -vv --exe smrf \
-    && python3 -m pip uninstall -y nose \
+    && python3 -m pip uninstall -y nose smrf[tests] \
```
jomey, line 39: I think that should make the Docker built happy again. :crossed_fingers:

scotthavens: This was trying to install smrf from pypi so it didn't. Updated the install to add the extras and the docker builds.

scotthavens (approved):

PR #196, opened 2020-11-02
Wind ninja interpolation

scotthavens:
If there are no nan values in the WindNinja data, don't interpolate. An error occurred when the topo resolution is the same resolution as WindNinja where there are no NaN value and an exception is raised in scipy.


jomey (approved):
If there is time, I would add a couple of unit tests around this. Seems straight forward to test

scotthavens:
> If there is time, I would add a couple of unit tests around this. Seems straight forward to test

I have a file now that is 50m for the Lakes, I'll add the test.

jomey:
> > If there is time, I would add a couple of unit tests around this. Seems straight forward to test
> 
> I have a file now that is 50m for the Lakes, I'll add the test.

Cool.
Some times I also just use `numpy.random` to test some basic logic around manipulating array data. Think this could work for this case too, no?

scotthavens:
Added a test with 50m WindNinja output. The implicit test is that there is no exception which was happening with the 50m output.

PR #201, opened 2021-01-21
SRMF data loading - Class structure

jomey:
Establishing a class inheritance structure for gridded input data to ensure some consistency across them.
This mainly takes care of initializing them and ensuring required attributes across them.


jomey:
This is the first result of looking into excluding wind data loading from HRRR, when katana is the chosen wind model. Breaking this up into an early PR since it seems a good first point to merge.

smrf/data/gridded_input.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @property
+    def config(self):
+        return self._config
+
+    @property
+    def topo(self):
+        return self._topo
+
+    def __set_attribute(self, attribute, argument):
```
scotthavens, line 33: Wouldn't this be easier to just remove the `None` default argument? Is this trying to make less code instead of specifying setters for each one?

smrf/data/hrrr_grib.py
```diff
@@ -1,14 +1,13 @@
-import logging
-
 import numpy as np
 import pandas as pd
 from weather_forecast_retrieval.hrrr import HRRR
 
 from smrf.envphys.solar.cloud import get_hrrr_cloud
 from smrf.envphys.vapor_pressure import rh2vp
+from .gridded_input import GriddedInput
```
scotthavens, line 7: Just my style, but I like having the full path to be verbose.

smrf/data/load_data.py
```diff
@@ -93,6 +67,38 @@ def __init__(self, smrf_config, start_date, end_date, topo):
         if self.data_type == InputCSV.DATA_TYPE:
             self.load_class.check_colocation()
 
+    def __determine_data_type(self, smrf_config):
```
scotthavens, line 70: Nice and clean!

smrf/tests/data/test_hrrr_grip.py
```diff
@@ -0,0 +1,30 @@
+import unittest
+
+import pandas as pd
+
+import smrf.data as smrf_data
+
+
+class TestInputGribHRRR(unittest.TestCase):
```
scotthavens, line 8: File name is `grip`, did you mean `grib`?

smrf/tests/smrf_test_case.py
```diff
@@ -52,6 +52,10 @@ def dist_variables(self):
     def base_config(self):
         return self.base_config_copy()
 
+    @property
+    def smrf_instance(self):
```
scotthavens, line 56: Good call, we do use the SMRF instance a lot! This is returning a new instance each time correct? I cannot be certain that there is some less than optimal coding that may change the SMRF instance, so I wouldn't want that to propagate to other tests.

scotthavens (commented):
I'll get the code running after our VPN is fixed to check out the code more in depth.

smrf/data/gridded_input.py
jomey: I went the conservative route by not changing the method signature. I agree that just removing 'None' would be a lot easier. Go for the bigger change then?

smrf/data/hrrr_grib.py
jomey: Thinking behind a relative path here is to hint that the base class is on the same level. I can go either way.

smrf/tests/data/test_hrrr_grip.py
jomey: Good catch!

smrf/tests/smrf_test_case.py
jomey: Correct, a new instance each time. Had the same thinking as you and wanted to ensure that there is a 'fresh' instance and less chance for test failures due to 'leftovers' from previous test cases.

smrf/data/gridded_input.py
jomey: Thinking about this a bit more, an advantage of this handling is that the error message will be more informative by giving the exact parameter that is missing. Removing the `None` would simply result in a "wrong number of arguments" error. Food for thought ...

scotthavens: Agree that it's more verbose the way you have it. I'm confused as to why the original was set to None to begin with then checked to make sure it wasn't None... I'll chock it up to learning python!

I'll leave it up to you, it seems that you have to have those parameters passed for the gridded data.

smrf/data/hrrr_grib.py
scotthavens: The `image_data` class is imported with the full [path](https://github.com/USDA-ARS-NWRC/smrf/blob/037d17a1b8920abb16cec90e3bc9ec339fbd1e23/smrf/distribute/air_temp.py#L1). Perhaps full path to keep it stylistically the same within SMRF but I understand the reasoning.

smrf/data/gridded_input.py
jomey: Let me go through the code one more time and see if the 'None` can be dropped. Do I hear that there are no other consumers of the class other than internal SMRF?

scotthavens: It should just be internal to SMRF, I don't know anything outside that uses it. Even AWSM wouldn't use it.

jomey: What do you think of this approach?

scotthavens: Looks good, gives more verbose errors when they happen.

jomey: I am ready to merge then ....

scotthavens: Trying to review it now, we're still having VPN issues...

smrf/data/load_data.py
```diff
@@ -64,26 +56,8 @@ def __init__(self, smrf_config, start_date, end_date, topo):
...
-                'config': self.smrf_config['gridded'],
-                'topo': self.topo
-            }
-
-        # load the data
-        self.load_class = self.DATA_FUNCTIONS[self.data_type](
-            self.start_date,
-            self.end_date,
-            **data_inputs)
+        self.__determine_data_type(smrf_config)
```
scotthavens, line 59: docstring?

smrf/tests/data/test_gridded_input.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def test_end_date(self):
+        hrrr_input = GriddedInput(**self.VALID_ARGS)
+
+        self.assertEqual(
+            self.END_DATE,
+            hrrr_input.end_date
+        )
+
+    def test_time_zone(self):
```
scotthavens, line 41: For completeness and since you've done a lot of work with timezones, does it make sense to add a single test for a non UTC timezone? While I hope if doesn't happen, if someone is using gridded input and specifies the start date for SMRF in MST then this might have issue. Or do we enforce a check in `model_framework` that one has to run SMRF in UTC with gridded data?

scotthavens (approved):
Looks good!

One thing I'm wondering outside of this PR is test documentation. I like how you've built a lot of the tests for the class setup and rely on the full suites to test that they work. But I'm wondering if someone was to just look at that, they might not know that?

smrf/data/load_data.py
jomey: Will add

smrf/tests/data/test_gridded_input.py
jomey: Good old timezones.

I think we are good for now. The arguments passed to the different input data classes are coming from the parsed `.ini`. If we were to enforce certain timezones, then I would vote to put that up higher in the stack. As you noted once earlier, be our guest if you want to run the model in non-UTC, but expect surprises.

This test mainly covers that the class is using the zone from the start date.

jomey:
> Looks good!
> 
> One thing I'm wondering outside of this PR is test documentation. I like how you've built a lot of the tests for the class setup and rely on the full suites to test that they work. But I'm wondering if someone was to just look at that, they might not know that?

I will add some documentation for the test suite.

smrf/tests/data/test_gridded_input.py
scotthavens: Timezones... gotta love them in life, hate them in programming!

Probably should add something to the read the docs and the README about timezones.

scotthavens:
So this will never pass Travis CI as we ran out of credits, I have requested more. @jomey let me know when you're ready and I'll use the big hammer. I have tested this locally and everything passes.

README.md
```diff
@@ -140,3 +140,15 @@ In a way that ARS uses this, we keep the config, topo and data on one location a
...
+
+### Tests
+SMRF relies on class level and integration type testing.
+On the class level, tests cover expected structure and ensure proper attribute types.
+Example: `tests/data/test_gridded_input.py`
+
+The integration tests ensure that changes to the code base do not trigger unexpected changes to the model output.
+Example: `tests/tes_full_smrf.py`
+
+All required data for integration tests is stored under: `tests/basins/` with two sample areas.
```
jomey, line 154: @scotthavens Does this go the direction you were thinking of?

smrf/data/gridded_input.py
```diff
@@ -0,0 +1,76 @@
...
+    def start_date(self, value):
+        if not isinstance(value, datetime):
+            raise TypeError('Argument start_date is not an instance of %s',
+                            datetime.__name__)
+        else:
+            self._start_date = value
+
+    @property
+    def time_zone(self):
+        return self.start_date.tzinfo
```
jomey, line 52: Snuck in one more change and not storing the `time_zone`, but reading it this way each time. For that to work I added a type check for the `start_date`, which should have been part of the first iteration.

smrf/data/input_data.py
```diff
@@ -4,9 +4,10 @@
 import utm
 
 from smrf.data import InputCSV, InputGribHRRR, InputNetcdf, InputWRF
+from smrf.data.gridded_input import GriddedInput
 
 
-class InputData():
+class InputData:
```
jomey, line 10: Note the rename of the file to reflect the class name.

README.md
scotthavens: Looks great. There is a typo in `test_full_smrf.py`

jomey:
> So this will never pass Travis CI as we ran out of credits, I have requested more. @jomey let me know when you're ready and I'll use the big hammer. I have tested this locally and everything passes.

Any luck with more credits?
Could we apply the big :hammer: to this PR?

scotthavens:
🔨

PR #206, opened 2021-04-27
Weather Forecast Retrieval - Update to v0.7

jomey:
Adapt the changes introduced in weather_forecast_retrieval v0.7.

### DIFF
--- a/smrf/data/hrrr_grib.py
+++ b/smrf/data/hrrr_grib.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pandas as pd
-from weather_forecast_retrieval.hrrr import HRRR
+import weather_forecast_retrieval.data.hrrr as hrrr_data
 
 from smrf.data.gridded_input import GriddedInput
 from smrf.envphys.solar.cloud import get_hrrr_cloud
@@ -55,14 +55,15 @@ class InputGribHRRR(GriddedInput):
             self.config['hrrr_directory']
         ))
 
-        metadata, data = HRRR(external_logger=self._logger).\
-            get_saved_data(
-                self.start_date,
-                self.end_date,
-                self.bbox,
-                output_dir=self.config['hrrr_directory'],
-                force_zone_number=self.topo.zone_number
-            )
+        metadata, data = hrrr_data.FileLoader(
+            file_dir=self.config['hrrr_directory'],
+            external_logger=self._logger
+        ).get_saved_data(
+            self.start_date,
+            self.end_date,
+            self.bbox,
+            force_zone_number=self.topo.zone_number
+        )
 
         self.parse_data(metadata, data)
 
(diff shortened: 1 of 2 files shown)


smrf/data/hrrr_grib.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 import pandas as pd
-from weather_forecast_retrieval.hrrr import HRRR
+import weather_forecast_retrieval.data.hrrr as hrrr_data
```
scotthavens, line 3: Why rename it? Could you not just import `FileLoader` or just do `hrrr.FileLoader`?

jomey: If you look at the call in line 58 and the import with reanming, this clearly indicates where the `FileLoader` is coming from. It basically combines the `hrrr.data` as `hrrr_data`. Using only `hrrr` is a very general name, has higher potential to be accidentally overwritten by a local variable, and using `hrrr_data` then also shows why and what this import does. Not importing the `FileLoader` also indicates in the lower line that this class is not part of SMRF and rather an imported class from WFR.

jomey: It follows the principle found for instance in the Google Python style guide, with the addition of renaming to prevent the above mentioned potential of error
https://google.github.io/styleguide/pyguide.html#224-decision

scotthavens: I understand why to rename it, I'm just not a fan as it seems to add an extra layer to the import. To me, this is then an bad name in WFR and should have been named something else to avoid someone using a duplicate variable name.

scotthavens (approved):

scotthavens:
Travis CI is off and forced the merge. This will be tested in the next PR where the Github actions are being setup.

smrf/data/hrrr_grib.py
jomey: I do not agree with calling this a bad name choice in WFR. Naming conventions should not only be driven on how other libraries would use it and also reflect the structure of the library itself. From that regard, the naming was a natural hierarchy to me.

Happy to remove this alias and use only `hrrr`. Then it is identical to the import as it was before. The alias was a conservative strategy based of observations in the current code base where variable names did not have a consistent descriptive choice.

scotthavens: I don't think it was a bad name choice either, it is the natural choice.

I think I'm more of a purist in this regard of not renaming packages, it's just a personal preference. I do see the benefit when they have a common name to rename the module. As SMRF/AWSM get to be more verbose, then this shouldn't be an issue.

jomey: Thought about some possible options to establish a common style used across the different libraries.
What do you think of this kind of import style:
```python
import weather_forecast_retrieval as wfr
```
Then the call becomes:
```python
wfr.data.hrrr.FileLoader
```

This convention leans on the numpy or matplotlib style, where there is a common acronym for both.

scotthavens: I like that, especially for WRF, it's such a long repo name.

jomey: Will update with a next PR

Issue #89, opened 2019-03-29
Marks2017 does not use storm total in calc_phase_density


micahjohnson150:
In precipitation.py distribute_for_Marks2017 calculates the density and phas for the snow like so:

https://github.com/USDA-ARS-NWRC/smrf/blob/master/smrf/distribute/precipitation.py#L341

`
                # determine the precip phase and den
                snow_den, perc_snow = snow.calc_phase_and_density(precip_temp,
                                                                  self.precip,
                                                                  nasde_model=self.nasde_model)
`

Which I believe is incorrect as I think it should be using the storm_total mass. like the following:

`
                # determine the precip phase and den
                snow_den, perc_snow = snow.calc_phase_and_density(precip_temp,
                                                                  self.storm_total,
                                                                  nasde_model=self.nasde_model)
`

I believe this was introduced here (May 2017 by me)  :(  

 https://github.com/USDA-ARS-NWRC/smrf/commit/7eebba6569bca8700a91fac936d7352578913d49#diff-b764dbb43f76c24e1c6c657401afea37R327

@micah-prime do you agree this is incorrect?

micah-prime:
I do agree that this is incorrect, but I don't think we can just swap the two. 

It looks like there are some operations that should stay hourly in the ```snow.marks2017``` function. For instance, we mask density and percent_snow by the pixels that have snow for each hour, and this would not work using the ```self.storm_total```. Maybe we need to pass hourly and storm total precip into the function? The issue here is that we would mess up the function definition for ```calc_phase_and_density```.

A work around could be to mask ```self.storm_total``` by the ```self.precip``` array. 
Thoughts @micahjohnson150 ?

PR #209, opened 2021-05-04
Wind Ninja - Skip loading wind data from HRRR

jomey:
Add logic that skips loading the wind data from HRRR once WindNinja is set as the wind model.

### DIFF
--- a/smrf/data/hrrr_grib.py
+++ b/smrf/data/hrrr_grib.py
@@ -1,8 +1,10 @@
 import numpy as np
 import pandas as pd
-import weather_forecast_retrieval.data.hrrr as hrrr_data
+import weather_forecast_retrieval as wfr
 
 from smrf.data.gridded_input import GriddedInput
+from smrf.distribute.wind import Wind
+from smrf.distribute.wind.wind_ninja import WindNinjaModel
 from smrf.envphys.solar.cloud import get_hrrr_cloud
 from smrf.envphys.vapor_pressure import rh2vp
 
@@ -15,9 +17,11 @@ class InputGribHRRR(GriddedInput):
         'air_temp',
         'vapor_pressure',
         'precip',
+        'cloud_factor'
+    ]
+    WIND_VARIABLES = [
         'wind_speed',
         'wind_direction',
-        'cloud_factor'
     ]
 
     TIME_STEP = pd.to_timedelta(20, 'minutes')
@@ -29,6 +33,17 @@ class InputGribHRRR(GriddedInput):
                 self.config['hrrr_load_method'] == 'timestep':
             self.timestep_dates()
             self.cf_memory = None
+
+        self._load_wind = not Wind.config_model_type(
+            kwargs['config'], WindNinjaModel.MODEL_TYPE
+        )
+
+    @property
+    def variables(self):
+        if self._load_wind:
+            return np.union1d(self.VARIABLES, self.WIND_VARIABLES)
+        else:
+            return self.VARIABLES
 
     def timestep_dates(self):
         self.end_date = self.start_date + self.TIME_STEP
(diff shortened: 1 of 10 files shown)


jomey:
The tests won't pass until https://github.com/USDA-ARS-NWRC/weather_forecast_retrieval/pull/33 will be merged

jomey:
Latest of WFR was picked up and tests pass now.

smrf/data/hrrr_grib.py
```diff
@@ -149,6 +160,34 @@ def parse_data(self, metadata, data):
...
+
+        Args:
+            data: Loaded data from weather_forecast_retrival
+        """
+        dataframe_options = dict(
+            index=data['air_temp'].index,
+            columns=data['air_temp'].columns
+        )
+        wind_speed = np.zeros_like(data['air_temp'].values)
+        wind_direction = np.zeros_like(data['air_temp'].values)
```
scotthavens, line 175: We've had a problem before of initializing with 0 values. 0 has meaning in SMRF and if something were to happen and not overwrite the dataframe, SMRF would continue on. The remedy was to initialize with NaN as that will cause SMRF to crash.

smrf/distribute/wind/wind.py
```diff
@@ -69,41 +70,52 @@ class Wind(image_data.image_data):
...
+            config (dict): run configuration for SMRF
             wind_model (str): name of the wind model
 
         Returns:
-            bool: True/False if the wind_model matches the config
+            bool: True/False if the wind_model is set in the config
         """
 
-        return self.config['wind_model'] == wind_model
+        return config.get('wind', {}).get('wind_model', None) == wind_model
```
scotthavens, line 118: Is the `.get('wind', {})` needed? inicheck will complain if the wind section is not in the config so it's safe to assume that `wind` will be there.

smrf/framework/model_framework.py
```diff
@@ -36,15 +36,15 @@
 from inicheck.config import UserConfig
 from inicheck.output import generate_config, print_config_report
 from inicheck.tools import check_config, get_user_config
+from smrf.envphys import sunang
+from smrf.utils import queue
```
scotthavens, line 40: Why move these? If you run `make isort`, it will probably move them back.

smrf/framework/model_framework.py
```diff
@@ -234,14 +234,14 @@ def loadTopo(self):
 
     def create_distribution(self):
         """
-        This initializes the distirbution classes based on the configFile
+        This initializes the distribution classes based on the configFile
```
scotthavens, line 237: Spelling is not my strong siut.

jomey: I let my IDE do that for me :wink:

smrf/framework/model_framework.py
jomey: That was the result of `isort`.
Running it with this version:
```shell
isort -c -v smrf/framework/model_framework.py
SUCCESS: /data/projects/ARS/smrf/smrf/framework/model_framework.py Everything Looks Good!
```

smrf/distribute/wind/wind.py
jomey: This will ensure that the call won't crash if you were to use SMRF outside of an initialized run via a `.ini`. 
It's something I adapted over the years, to never assume a root key exists when checking for a child key.
The existence check for a key is easy to do and does not hurt performance

smrf/framework/model_framework.py
scotthavens: Well then, that is odd.

smrf/framework/model_framework.py
scotthavens: This was pre good IDE's when that was written...

smrf/framework/model_framework.py
jomey: Was also surprised to see that as part of the commit

scotthavens (approved):

scotthavens:
Does this save significant time when loading HRRR files? This PR is the source of the error in AWSM PR [#89](https://github.com/USDA-ARS-NWRC/awsm/pull/89). When the data is loaded with threading and HRRR is loaded for each time step, then the `data_queue` doesn't have the correct variable names.

jomey:
> the `data_queue` doesn't have the correct variable names.

Can I get more context?
What would be the correct variable names? Why do the SMRF tests pass, but not the the AWSM?
Wouldn't the proper solution be to fix/create the correct names instead of rolling this change back?

scotthavens:
See #210

PR #210, opened 2021-05-25
HRRR data queue

scotthavens:
- had to add a `data_variables` to the hrrr_grib to load the correct variables into the data queue
- Add new tests for the Lakes to test different methods of running HRRR with loading first or for each timestep
- tests using the mixin to check netcdf files

### DIFF
--- a/smrf/data/hrrr_grib.py
+++ b/smrf/data/hrrr_grib.py
@@ -41,9 +41,13 @@ class InputGribHRRR(GriddedInput):
     @property
     def variables(self):
         if self._load_wind:
-            return np.union1d(self.VARIABLES, self.WIND_VARIABLES)
+            return self.data_variables
         else:
             return self.VARIABLES
+
+    @property
+    def data_variables(self):
+        return np.union1d(self.VARIABLES, self.WIND_VARIABLES)
 
     def timestep_dates(self):
         self.end_date = self.start_date + self.TIME_STEP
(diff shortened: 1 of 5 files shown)


smrf/data/hrrr_grib.py
```diff
@@ -111,7 +115,7 @@ def load_timestep_thread(self, date_times, data_queue):
         for date_time in date_times[1:]:
             self.load_timestep(date_time)
 
-            for variable in self.variables:
+            for variable in self.data_variables:
```
scotthavens, line 118: This was the major issue. The `variables` were filtered to non-wind variables where the `data_queue` was expecting there to be values to get (even if they aren't used).

scotthavens:
Testing locally, this fixes the errors seen in AWSM PR [#89](https://github.com/USDA-ARS-NWRC/awsm/pull/89)

smrf/data/hrrr_grib.py
jomey: Do I understand this correct that the threaded solution will revert back to loading the data even though not used?

scotthavens: It's not loading wind so it keeps your PR #209. On line 78 it still filters out the wind variables so they aren't loaded. This just ensures that the `data_queue` is populated using the empty dataframes set in `InputGribHRRR.calculate_wind`

jomey: Got it. Mainly asking to make sure there is consistent behavior independent of running threaded or not.

jomey (approved):

PR #213, opened 2021-07-27
Precip - input and output data scalars and testing

micah-prime:
Precip - methods and config options to scale precip pre and post distribution. These methods can be implemented for other variables and can be extended to add more approaches to data scaling. Additionally make a test case to test config permutations and evaluate result on small array in memory. Add the parameterized package as a dev requirement to allow parameterized tests within unittest.TestCase


micah-prime:
@scotthavens @jomey hoping you can take a look at this when you have some time.

The idea is to allow modifications of input data both pre and post distribution to account for known biases in the source of the data. I started with a simple case (just scale the data by a factor) in order to get the code scaffolded.

In order to test this without creating another set of test files, I added a test case that uses a smaller mocked DEM and can test the expected post-distribution numpy array for a variable.

Open to any feedback. I'll add a few comments through the PR of specific places I'd love some input on.

smrf/framework/model_framework.py
```diff
@@ -48,6 +48,22 @@
 from smrf.utils.utils import backup_input, date_range, getqotw
 
 
+class SolarParams:
```
micah-prime: I added this class to track the solar parameters within a timestep. Any thoughts on where it should live?

smrf/tests/mock_helpers.py
```diff
@@ -0,0 +1,57 @@
+from collections import OrderedDict
+
+import numpy as np
+
+import smrf
+
+
+class MockTopo(smrf.data.load_topo.Topo):
```
micah-prime: This is the upper 4x4 grid from the RME topo. I liked the idea of having the topo mocked out for the smaller tests.

smrf/tests/test_distribute_permutations.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ConfigPermutations:
+    def __init__(self, config_changes, distribute_section, variable, expected_array):
+        self.config_changes = config_changes
+        self.distribute_section = distribute_section
+        self.variable = variable
+        self.expected_array = expected_array
+
+
+class TestDistributePermutations(SMRFTestCase):
```
micah-prime: The idea here is that we can track that an expected change is happening to a variable distribution based on the changes in the config.

smrf/distribute/precipitation.py
```diff
@@ -282,6 +284,12 @@ def distribute(self, data, dpt, precip_temp, ta, time, wind, temp,
                         temp,
                         self.config,
                         self.metadata)
+                # Scale input data based on config
+                if input_scalar_key is not None:
```
scotthavens: Since this is being applied to all the input data, why not do this when it's calculated for corrected precip? Indexing with pandas can be quite slow.

smrf/distribute/precipitation.py
```diff
@@ -301,10 +309,20 @@ def distribute(self, data, dpt, precip_temp, ta, time, wind, temp,
...
+                    data.precip = scale_data(
```
scotthavens: Similar concept here. This is being loaded first so could just multiply the dataframe when it's loaded by the scalar instead of each time step.

smrf/distribute/precipitation.py
```diff
@@ -301,10 +309,20 @@ def distribute(self, data, dpt, precip_temp, ta, time, wind, temp,
...
+                        data.precip, input_scalar_key, self.config)
 
                     self.distribute_for_susong1999(
                         data, precip_temp, time)
         else:
+            # Scale input data based on config
+            if input_scalar_key is not None:
```
scotthavens: This is doing the same as the above, correct? Suggest this is all moved to when the data is loaded.

smrf/distribute/precipitation.py
```diff
@@ -324,6 +342,13 @@ def distribute(self, data, dpt, precip_temp, ta, time, wind, temp,
                     self.veg,
                     self.config)
 
+        # scale the post distribution data based on user config
+        output_scalar_key = get_scalar_value_key(self.config, "output")
+        if output_scalar_key is not None:
```
scotthavens: Thinking like @jomey, could this be moved to a `@property`? There might need to be a setter/getter but if the scaling was desired, then the property could change the distributed value and return.

smrf/framework/CoreConfig.ini
```diff
@@ -649,6 +649,27 @@ default = us_nws_8_shielded,
...
+
+input_scalar_factor:
+default = 1.0,
+type = float,
+description = Scale the input data by a constant factor
+
+output_scalar_type:
+default = None,
+options = [None factor],
+description = Method for scaling data after distribution
```
scotthavens: Might want to specify that this is only being done for `precip` and not anything else like new snow density.

smrf/framework/model_framework.py
scotthavens: Probably in `utils` until we have a more complicated solar module. Another option is to used a `namedtuple` and just initialize it in the `_prep_solar` method. Unless there is thoughts that methods would be added at some pont?

https://docs.python.org/3/library/collections.html#collections.namedtuple

smrf/framework/model_framework.py
```diff
@@ -401,14 +417,14 @@ def disttribute_data_serial(self):
...
 
-    def distribute_single_timestep(self, t):
-
-        self._logger.info('Distributing time step {}'.format(t))
-
-        if self.hrrr_data_timestep:
-            self.data.load_class.load_timestep(t)
-            self.data.set_variables()
-
+    def _prep_solar(self, t):
```
scotthavens: Can we rename? The method name isn't really reflecting what is being done which is sun angle. Even `solar_parameters` like the output variable would be descriptive.

smrf/framework/model_framework.py
```diff
@@ -423,6 +439,17 @@ def distribute_single_timestep(self, t):
                 self.topo.aspect,
                 azimuth,
                 cosz)
+        return SolarParams(cosz, azimuth, rad_vec, illum_ang)
```
scotthavens, line 426: The `namedtuple` could live here with a lot less code than the class if you go that route.

smrf/framework/CoreConfig.ini
```diff
@@ -649,6 +649,27 @@ default = us_nws_8_shielded,
 options = [us_nws_8_shielded us_nws_8_unshielded],
 description = WMO model used to adjust for undercatch of precipitaiton
 
+input_scalar_type:
+default = None,
+options = [None factor],
```
scotthavens, line 654: Should have a recipe with this for when scalar type is None, then the factor is removed.

smrf/data/modifiers.py
```diff
@@ -0,0 +1,43 @@
+def get_scalar_value_key(config, stage):
```
scotthavens: Wondering if `stage` is descriptive enough. It's the stage at which the scalar is applied but could there be a better term?

smrf/data/modifiers.py
```diff
@@ -0,0 +1,43 @@
...
+        raise ValueError(
+            f"{stage} is not an allowed stage value for data scaling"
+        )
+    method = config.get(f"{stage}_scalar_type")
+    if method is None:
+        return None
+    return f"{stage}_scalar_{method}"
+
+
+def scale_data(data, scalar_key, config):
```
scotthavens: Could this not just take data and scalar as an argument? Seems a bit much to pass the config section just to extract a config value.

setup.py
```diff
@@ -147,6 +147,7 @@ def finalize_options(self):
         ],
         'tests': [
             'mock',
+            'parameterized~=0.8.1',
```
scotthavens, line 150: Love this feature for pytest, glad there is a unittest equivalent!

smrf/tests/data/test_modifiers.py
```diff
@@ -0,0 +1,57 @@
...
+        ]
+    )
+    def test_get_scalar_value_key(self, stage, config, expected):
+        assert modifiers.get_scalar_value_key(config, stage) == expected
+
+    def test_get_scalar_value_key_failure(self):
+        with self.assertRaises(ValueError):
+            modifiers.get_scalar_value_key({}, "bad_stage")
+
+    @parameterized.expand(
```
scotthavens, line 47: Nice! Gonna be using this one!

smrf/tests/test_distribute_permutations.py
scotthavens: Would a simpler approach be to test that the output was scaled compared to gold? So instead of creating a mock topo (which adds another level of complexity) test that the new precip field divided by the gold field which would equal the scale factor for the entire array?

Then you could test a bunch of different parameterizations to both the inputs and outputs without having to redefine the expected results each time. As it is, this would really only test that the output data was scaled but not if the input data was scaled.

This goes a bit against testing methodology since it wouldn't test the output directly but assure that the factor was applied.

scotthavens (changes requested):

scotthavens:
Thanks @micah-prime  for the updates, seems like the CO modeling could have used this last year! Overall it's a great contribution, many of my comments are aimed at fit a bit more into the design/style of SMRF.

micah-prime:
@scotthavens thanks for the review! CO was definitely the inspiration for these changes. 

I'll address the comments shortly :)

smrf/framework/model_framework.py
jomey: I could also see this just moving to the `utils`. Namedtuples have more features than you need and I would go with simple.

My new favorite for Python 3.7 and above are data classes for objects that are type aware:
https://docs.python.org/3/library/dataclasses.html

Maybe one day, once SRMF drops 3.6 support

smrf/tests/test_distribute_permutations.py
jomey: I am also torn whether the level of complexity of the Mock is balanced with the benefit of the test.
Could you add a unit level test on the `precipitation` class and forgo the integration level test for the parameter? Or as Scott suggested, the integration level test mainly checks that the outputs compared to the Gold just be larger or smaller depending on the scale factor.

smrf/data/modifiers.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if scalar_key in ["input_scalar_factor", "output_scalar_factor"]:
+        scalar = config[scalar_key]
+        scaled_data = data * scalar
+    else:
+        raise NotImplementedError(
+            f"{scalar_key} is not an implemented scalar method."
+        )
+
+    return scaled_data
```
jomey: Minor personal preference: Would move this return statement inside the `if` block.

smrf/tests/data/test_modifiers.py
```diff
@@ -0,0 +1,57 @@
...
+
+class TestModifiers(unittest.TestCase):
+    @parameterized.expand(
+        [
+            ('input', {}, None),
+            ('input', {'input_scalar_type': 'factor'}, 'input_scalar_factor'),
+            ('output', {'output_scalar_type': 'factor'},
+             'output_scalar_factor'),
+            ('output', {}, None)
+        ]
```
jomey, line 35: How about a scenario with unknown keys to test error cases?

smrf/data/modifiers.py
jomey: Personal preference: As a person who likes OOP - I would make this a class `Modifiers` and make the two methods `@classmethod`. Then you would import that class and your code becomes a bit more descriptive at the places you call those methods.

jomey (commented):
Thanks Micah for sharing the code. You couldn’t have timed this better as I am running into the same challenge with precip in the ERW runs. Wondering how this parameter plays together with the `detrend` turned off?
I am currently experimenting with that as well

smrf/framework/model_framework.py
micah-prime: `utils` sounds good. Yeah, data classes are my preference for something like this as well. Some day!

smrf/tests/test_distribute_permutations.py
micah-prime: I'm torn here as well. My initial thought was that we may want to add more complicated scalar methods down the line, like maybe an approach that scales the precip by elevation if that is the systematic error we know exists in the input data. With a more complicated scaling approach, I'm inclined to stay away from the gold files and lean into a 4x4 array that we can specify in the test.

My current approach does feel like a mix and match of integration and unit testing, though. Maybe a pure unit test like @jomey suggested with all the data mocked out would be a better tact. This could be a better testing paradigm for testing more permutations of the config.

Thoughts?

smrf/distribute/precipitation.py
micah-prime: Good catch! My thought process was that we wanted to perform undercatch correction before scaling, but looking back at the undercatch correction methods, the order won't make a difference. This was an oversight on my part.

smrf/distribute/precipitation.py
micah-prime: I like this idea a lot. Are you suggesting the `output_scalar_key` be the property, or we make a property for something like `scaled_data`?

Or should I make this a property on the `Modifiers` class that @jomey suggested a couple comments down?

smrf/data/modifiers.py
micah-prime: @scotthavens yeah I bet there is a more descriptive term. What if instead of `stage` it was a boolean for `pre_distribution` since it is either pre or post distribution? 

@jomey great call. I'll make that change.

smrf/data/modifiers.py
micah-prime: This can definitely be simplified. I wanted to leave room for more scaling methods down the line, but I can definitely avoid passing the config here. Making the `Modifiers` class will also help with this.

setup.py
micah-prime: Yeah! I was pleasantly surprised to find this.

smrf/data/modifiers.py
micah-prime: Can do!

smrf/tests/data/test_modifiers.py
micah-prime: Good call, will do!

smrf/data/modifiers.py
jomey: With going the class route, this could also be less dynamic and simply have two methods there. One for the input and one for the output key. Both key names can then be set as constants on the class and also used below on line#35 or have that also be replaced as a class constant to hold scalar keys options.
My idea would look like this:
```python
class Modifiers:
    INPUT_SCALAR_KEY = "input_scalar_factor"
    OUTPUT_SCALAR_KEY = "output_scalar_factor"
    KEY_OPTIONS = [INTPUT_SCALAR_KEY, OUTPUT_SCALAR_KEY]

    # two methods for the key retrieval
```

smrf/framework/model_framework.py
scotthavens: Dataclasses are neat! Can't quite drop 3.6 support yet but EOL is slated for 12/2021.

smrf/tests/test_distribute_permutations.py
scotthavens: I can see this being added upon as well in the future so we should think about the best implementation for this.

I'm a big fan of integration testing, especially when we are talking about modifying outputs. Then this needs to be tested again the gold files. While I see the benefit of having a smaller subet to work with so you can actually see the results, a small domain does now capture a large enough sample size. We have already seen it with the RME test case where it cannot pick up issues with the code because it's too small, which is why we moved to the Lakes. If it takes even 1 more minute to run the tests then that is an okay tradeoff.

With that said, I agree that a few more unit tests would be useful. In my opinion, the integration test should test against the gold file in the Lakes to ensure that we back out the correct precipitation field. This will ensure that nothing else is being introduced from the scaling factors.

smrf/distribute/precipitation.py
scotthavens: If this is gridded data, SMRF should not be doing undercatch correction since it's not measured data. For measured data, it shouldn't matter since you could apply the factor outside SMRF (to the csv files) as well.

smrf/distribute/precipitation.py
scotthavens: I was thinking more of the `scaled_data` as a property. So it could even live on `image_data` at some point if this is applied to all the variables.

With it being on `Modifiers`, then you'd have to initiate a class and call the method anyways. Wouldn't that have to still live in the `@property` regardless?

smrf/data/modifiers.py
scotthavens: The modifiers can only be applied in to places for the input and output, right? Lets just keep it verbose to that, `pre_distribution` makes an assumption on where the factor will be applied.

Building of @jomey  class style, would there be explict methods for inputs and outputs? We are already going to have to call these anyways in specific spots in the input/output code. Something along the lines below where `config` could be passed on `__init__` of the class and you explicitly ask to scale the input/output using the config and keys from above. There would be similar code for the outputs as well. It makes for a bit more code but a lot more understandable as to when the factors are being applied.

```python
def get_input_value(self):
    return self.config.get(self.INPUT_SCALAR_KEY)

def scale_input(self, data):
    return data * self.config.get(self.INPUT_SCALAR_KEY, None)
````

Comment (micah-prime) [reply smrf/data/modifiers.py]: I'm a big fan of the separate methods now that it was brought up. I'm just about done with a refactor to the class that should make it more verbose and make it extensible for other scaling approaches later on. I'll be pushing later today once I make the integration test changes.

File: smrf/tests/test_distribute_permutations.py (hunk 3 above)
Comment (micah-prime) [reply smrf/tests/test_distribute_permutations.py]: @scotthavens I refactored the test to be an integration test, checking the ratio of the gold file to the test file. It is currently setup with the RME test case. I'm happy to swap it over to use Lakes, but the Lakes tests fail for me locally (maybe a Mac thing?) so I'm sticking with RME until the test methodology is approved.

File: smrf/distribute/precipitation.py (hunk 4 above)
Comment (micah-prime) [reply smrf/distribute/precipitation.py]: This should be addressed by the latest round of commits.

File: smrf/distribute/precipitation.py (hunk 7 above)
Comment (micah-prime) [reply smrf/distribute/precipitation.py]: The latests changes has a `Modifiers` class that is a `@property` in the precip class. It would be easy to transfer to the `image_data` class when we're ready for that.

File: smrf/tests/test_distribute_permutations.py (hunk 3 above)
Comment (scotthavens) [reply smrf/tests/test_distribute_permutations.py]: @jomey implemented an environment variable `NOT_ON_GOLD_HOST` which will relax the test requirements. It should pass on the CI actions when running with Ubuntu.

Comment (micah-prime) [reply smrf/tests/test_distribute_permutations.py]: I had to brew install eccodes. That created a sneaky runtime error that didn't show up in the logs at all. About to push again with the test changed to use the Lakes data

File: smrf/framework/CoreConfig.ini (hunk 11 above)
Comment (micah-prime) [reply smrf/framework/CoreConfig.ini:654]: Missed this in my first read through. Will address in the next commit

**Comment by micah-prime** [comment]
Thanks for the input on this PR! I believe I've addressed all of the comments so far. Let me know what you think!

**Review by jomey — APPROVED** [review]
Thanks Micah. Think this is a great start introducing scalar factors. Can see this being expanded to any input data adjustments that might come up.

File: smrf/distribute/precipitation.py [hunk 18]
```diff
@@ -194,6 +196,8 @@ def initialize(self, topo, data, date_time=None):
                 data.precip,
                 self.storms,
                 stations=self.stations)
+            self.corrected_precip = self.modifier.scale_input_data(
```
scotthavens: Following the logic that this should scale the input data, wouldn't this be applied in the wrong spot? This isn't input data but processed input data for this particular new snow density model.

smrf/distribute/precipitation.py
```diff
@@ -268,6 +278,7 @@ def distribute(self, data, dpt, precip_temp, ta, time, wind, temp,
 
         self._logger.debug('%s Distributing all precip' % data.name)
         data = data[self.stations]
+        data = self.modifier.scale_input_data(data)
```
scotthavens: This should be applied when the data is loaded. Running this on each timestep seems a bit overkill. If this is going to be applied to other variables down the road, then having it in `data` is more logical.

Also, while I'm sure multiplying by a factor isn't computationally expensive, if you're using `marks2017`, then this is being ran twice.

smrf/tests/__init__.py
```diff
@@ -1 +1,3 @@
 # -*- coding: utf-8 -*-
+
+non_distributed_variables = ['time', 'x', 'y', 'projection']
```
scotthavens: @jomey should this be uppercase to reflect it's a constant variable?

smrf/tests/check_mixin.py
```diff
@@ -36,3 +38,61 @@ def test_snow_density(self):
...
+        TestSomethingNew(CheckSMRFOutputRatios, SMRFTestCase)
+            RATIO_MAP = {
+                'air_temp': 1.0,
+                'precip': 2.0,
+            }
+    """
+    RATIO_MAP = {}
+    DECIMAL = 4
+
+    def _assert_ratio(self, file_name, variable):
```
scotthavens, line 58: Any particular reason for the underscore?

smrf/tests/test_distribute_permutations.py
```diff
@@ -0,0 +1,35 @@
+from inicheck.tools import cast_all_variables
+
+from smrf.framework.model_framework import run_smrf
+from smrf.tests.smrf_test_case_lakes import SMRFTestCaseLakes
+from smrf.tests.check_mixin import CheckSMRFOutputRatios
+
+
+class TestDistributePermutations(CheckSMRFOutputRatios, SMRFTestCaseLakes):
```
scotthavens: This might be better located in `tests/distirbute`

smrf/tests/test_distribute_permutations.py
```diff
@@ -0,0 +1,35 @@
...
+        super().setUpClass()
+
+        cfg = cls.configure_permutation({"precip": {
+            "input_scalar_type": "factor",
+            "input_scalar_factor": 1.0,
+            "output_scalar_type": "factor",
+            "output_scalar_factor": 2.0
+        }})
+
+        run_smrf(cfg)
```
scotthavens: This is really only testing the output factor correct?

Testing the input change to the gold file will be a lot harder to validate as it'll change the lapse rate that is calculated. But we should be aware that this aspect of the modifier component is not tested. If the input ratio was moved to the `data` class as mentioned above, then it would be easier to test.

smrf/tests/smrf_test_case.py
```diff
@@ -216,3 +221,33 @@ def compare_netcdf_files(self, output_file):
 
         gold.close()
         test.close()
+
+    def evaluate_netcdf_ratio(self, output_file):
```
scotthavens: Would this ever be used outside the `CheckSMRFOutputRatios` mixin? Suggest moving to the mixin since that is what is really using this method.

smrf/framework/recipes.ini
```diff
@@ -69,6 +69,20 @@ trigger_section:
 precip:
   apply_defaults = True
 
+[precip_input_scalar_recipe]
```
scotthavens, line 72: I've been trying to test recipes as well to ensure that they are working. I have only implemented it for wind but it's something we're tying to build out since there are some odd things that can happen with the recipes and we want to make sure that we're not getting any extra outputs that we don't need.

scotthavens (changes requested):

smrf/tests/__init__.py
jomey: I would stick with how the variables are capitalized in the netCDF

smrf/tests/check_mixin.py
jomey: Methods with an underscore in Python are considered private to the class. I agree with the use here since the ‘public’ ones are testing the variables and have a helper method internal to the class.

smrf/framework/recipes.ini
micah-prime: Sounds good. I'll take a look and see what I can do

smrf/tests/__init__.py
jomey: Reading this a second time, I think I misunderstood your comment.

Yes - to capitalization and also could have a better home in the
`SMRFTestCase` base class since we are defining a similar constant on line#25 there

smrf/tests/test_distribute_permutations.py
micah-prime: The root level test probably isn't the best spot. I was sticking with convention of the other integration level tests living there. Long term we might think about a `tests/integration` directory.

smrf/tests/smrf_test_case.py
micah-prime: Likely not. I'll move it there!

smrf/tests/__init__.py
micah-prime: Good advice all around. I'll move it and capitalize it.

smrf/tests/test_distribute_permutations.py
scotthavens: We don't really separate out unit and integration tests well. I've been trying to follow keeping tests in the same directory structure as SMRF, whether that's the best way to do it or not!

smrf/framework/recipes.ini
scotthavens: The recipe tests are fairly straight forward, there is an example in there already to follow but this will be a lot simpler.

smrf/distribute/precipitation.py
micah-prime: It is being run twice if `marks2017` is used, but on two separate datasets so it will not be compounding. 

I'm not sure how to implement this in `data` without getting into a specific implementation for each of the `InputData.load_class` options. If I understand the procedure correctly, the hrrr data timesteps are read in at runtime for each distribution timestep. https://github.com/USDA-ARS-NWRC/smrf/blob/master/smrf/framework/model_framework.py#L409

Running this operation on each timestep is not ideal, but the alternative was having unique logic for station and hrrr data, so I leaned towards the former. I can switch it if you want though, I'm not stuck on this approach. 
One option is scaling it here https://github.com/USDA-ARS-NWRC/smrf/blob/master/smrf/framework/model_framework.py#L330
and then a loop to scale input data by variable after this line NWRC/smrf/blob/master/smrf/framework/model_framework.py#L409

Or I could pass a `Modifier` object into the `InputData` class and implement a `load` method on the `InputData` class that enforces that the input data is modified. This would mean moving the `modifier` property to image data now and it just wouldn't do anything for the other variables. 

What is your preference?

scotthavens: Hmmm, forgot about the loading by timestep...

Building off the last option, would be to move scaling to the `InputData` class. Both stations and gridded use it to load the data (`load_class.load()`). After the data is loaded with any method, the `set_variables` method is called which takes the variables and adds them to the `InputData` instance. This could be one possible location for the scaling. So instead of `set_variables` it could be `set_and_scale_variables` or something like that. Within the method, it could scale the data.

https://github.com/USDA-ARS-NWRC/smrf/blob/465d42cca85820e76a50bc311d101c7dc506df8c/smrf/data/input_data.py#L116

This has a few benefits. First is it's in one location and second, makes it easily scalable to any other variable when it's needed. I don't think this would need to be on the image data as it'll be fully contained with the input data.

micah-prime: I do like that as a place for the input scaling. That means the modifiers will need to be initialized within the `InputData` class as well, but that doesn't seem like a bad thing.

smrf/framework/recipes.ini
micah-prime: This should be addressed in the latest round of commits.

smrf/tests/test_distribute_permutations.py
micah-prime: Yep! I added tests in the latest round of changes.

smrf/distribute/precipitation.py
micah-prime: The code I just pushed accomplishes this.

smrf/data/input_data.py
```diff
@@ -58,9 +69,15 @@ def __init__(self, smrf_config, start_date, end_date, topo):
 
         self.__determine_data_type(smrf_config)
 
+        # create a dictionary mapping variable name to modifier
```
scotthavens, line 72: Awesome, that will make it easy to modify other variables easily if needed.

micah-prime:
@scotthavens how's this looking to you?

micah-prime:
After chatting with @Hedrick-ARS it sounds like ARS did not think this is the right repo for this feature. I do still think this fits well with the goals of SMRF, but if the consensus is that it does not belong here, do you want me to close the PR?

scotthavens:
Sorry for the silence. I don't have much time to work on this anymore.

To document the reasons for the decision. The decision that this was not a good fit to be directly in SMRF is because the objectives of SMRF is to focus on the models for taking the weather data and creating inputs for iSnobal. In all of the development and simulations that we have performed using SMRF, any adjustments to the weather data was done outside of SMRF. This also follows the pattern of how IPW was built, none of the IPW utilities modify the inputs before calculating the variables.

With CSV files, this is a fairly straightforward problem by modifying the CSV files for different scaling factors and is something we have done many times. Gridded data provides a different problem due to the size of the input data. This would fit well in a SMRF pre-processor that could be added perhaps to `weather_forecast_retrieval` to create ne w files, etc.

micah-prime:
@scotthavens I appreciate the response. Hope things are going well!

Issue #214, opened 2021-11-11
Unknown veg_ values cause issues in albedo decay methods


micah-prime:
Unknown veg_ values in the config parse as a list of strings. This means that the date_method of albedo decay will fail when running with unknown veg_ config options

scotthavens:
To be clear, this is happening when you add a `veg_` option that isn't present in the `topo.nc`?

While this may be a pain for the operational release, is it worth the fix? Seems like the easier solution is to fix the configuration file?

PR #215, opened 2021-11-11
Unknown veg_ values cause issues in albedo decay methods

micah-prime:
* add test showing that albedo config parses unknown veg_ values as a list of strings rather than a float. 
* Update requirements for breaking build. use_2to3 is not allowed with latest setuptools
* Address issue #214

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,7 +23,7 @@ oset==0.1.3
 pandas==0.22.0
 pbr==3.1.1
 progressbar2==3.6.2
-pybtex==0.21
+pybtex==0.24
 pybtex-docutils==0.2.1
 pycodestyle==2.3.1
 pyflakes==1.5.0
--- a/tests/test_configurations.py
+++ b/tests/test_configurations.py
@@ -1,8 +1,10 @@
 import unittest
 import os, shutil
-from inicheck.tools import get_user_config, check_config
+from inicheck.tools import get_user_config, check_config, cast_all_variables
+from copy import deepcopy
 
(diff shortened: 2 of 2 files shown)


micah-prime:
@Hedrick-ARS @jomey @scotthavens
Could you give this a look?
This first commit should show where the code breaks, then I will push the fix. 

Also, does release-0.9 still build?

requirements.txt
```diff
@@ -23,7 +23,7 @@ oset==0.1.3
 pandas==0.22.0
 pbr==3.1.1
 progressbar2==3.6.2
-pybtex==0.21
+pybtex==0.24
```
micah-prime, line 26: version 0.21 won't build with the latest setup tools due to `use_2to3` being a deprecated option.

jomey:
Don't have enough insights for the 0.9 release.
The code base back then also was tested against Travis and that has since changed.

What is the motivation to have this only changed in an older release?

micah-prime:
@jomey it should be fixed in the newer release as well. I can take a look at that soon. I just started with the stable version.

requirements.txt
scotthavens: Why is this package needed and where is it used? The current version removed a lot of requirements that aren't needed.

scotthavens:
We ran out of credits on Travis and they won't give us more so we moved to GH actions. This won't build anymore so you'll have to limp this along until a new release can be cut.

Let's discuss this in #214 as I don't see the reason for the fix yet.

